PR #7, opened 2017-08-24
Kcu1500

ruck314:
1) updating BAR0 base address mapping
2) removing c++ prom load support from data_dev because it is now done in pyrogue


ruck314:
Required changes such that we can do do the PCIe card PROM update via pyrgoue (instead of non-common c++ driver application code).  Here's an example of how I am able to program the pgp-gen4-pcie-card's two QSPI PROMs via pyrogue:
https://github.com/slaclab/pgp-gen4-pcie-card/blob/master/software/scripts/updateStaticImageProm.py

ruck314:
this branch is now stable and working.  Merging it with master.

PR #32, opened 2019-01-16
Allow 64-bit buffers for streaming mode

slacrherbst:
Changes buffer allocation flags to allow 64-bit buffers in streaming mode.


weninc:
looks good to me

PR #38, opened 2019-05-20
Remote unused alignment check

slacrherbst:
The unused alignment check was causing problems running on 32-bit arm.


slacrherbst:
https://github.com/slaclab/aes-stream-drivers/releases/tag/v5.5.0

PR #51, opened 2019-09-05
PCI Enable Master Mode Fix

slacrherbst:


common/driver/axis_gen2.c
```diff
@@ -536,6 +536,7 @@ void AxisG2_SeqShow(struct seq_file *s, struct DmaDevice *dev) {
    seq_printf(s,"     Hw Read Buff Count : %i\n",hwData->hwRdBuffCnt);
    seq_printf(s,"           Cache Config : 0x%x\n",(ioread32(&(reg->cacheConfig))));
    seq_printf(s,"            Desc 128 En : %i\n",hwData->desc128En);
+   seq_printf(s,"            Envable Var : 0x%x\n",(ioread32(&(reg->enableVer))));
```
ruck314: Is it possible for the enableVer to be 0x0 if the driver is loaded?

slacrherbst: Yes if it is a older version of the firmware.

slacrherbst: But no the driver should enable it.

slacrherbst: But that register also has other useful information.

PR #57, opened 2020-06-12
Add irq holdoff to dma driver

slacrherbst:


ruck314:
@slacrherbst Where does the kernel driver case on Version >= 0x3 to know that the `dev->cfgIrqHold` new feature is supported?

slacrherbst:
I added the check. I don't understand why so many files show as changed when pulling to pre-release.

PR #58, opened 2020-07-08
Add support for buffer groups

slacrherbst:
Also resolve compile issues with latest kernel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,20 +29,8 @@ addons:
 stages:
     - name: gen_release
       if: tag IS present
-    - name: deploy_tag
+    - name: deploy_dkms
       if: tag IS present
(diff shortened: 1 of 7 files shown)


ruck314:
Confirmed legacy FW and new driver still compatible with each other

ruck314:
I think `hwData->bgEnable` should be initialized to 0x0 in `AxisG2_Init()` before line#388 (else it is uninitialized memory and possible to be 0x1. 

Here's how I think it should get init to 0x0:
```
   // Setup buffer groups
   hwData->bgEnable = 0x0
   if ( ((ioread32(&(reg->enableVer)) >> 24) & 0xFF) >= 4 ) {
...
   }
```

slacrherbst:
I thought I had a memset but I don't I will init the whole hardware record earlier.

PR #61, opened 2020-07-10
Fix irq buffer return

slacrherbst:


ruck314:
Looks good.  No longer missing buffers if closing DMA and buffers in the pipeline and no longer locking up if DMA closed and FW streaming

PR #84, opened 2022-03-08
ref count to prevent premature driver unload

LaVillaStargato:
ref count to prevent premature driver unload. kernel has this facility, here we use it.


LaVillaStargato:
ref count to prevent premature driver unload.  kernel has this facility, here we use it.

bengineerd (approved):

PR #91, opened 2022-05-16
Move data processing into work queue and support pure polling mode

slacrherbst:
This PR has two key changes.

The first change is to move most of the DMA handling out of the IRQ context and into a work queue thread which is queued in the interrupt processor. The IRQ handler will disable interrupts in the card and add a processing function onto the work queue. This function will then process the dma buffers and re-enable interrupts upon completion.

The second change is to support a pure polling mode where a polling function is repeatedly placed on a work queue in order to monitor the state of the dma buffers. In this mode interrupts on the card are never enabled.

I also removed a kmalloc and kfree which were occurring in the irq processing routine.

I am posting this PR for initial review. I will also need to perform high rate tests in both IRQ and polling mode to see how this impacts 1Mhz operation.

### DIFF
--- a/common/driver/axis_gen2.c
+++ b/common/driver/axis_gen2.c
@@ -169,7 +169,6 @@ uint32_t AxisG2_Process (struct DmaDevice * dev, struct AxisG2Reg *reg, struct A
          }
       }
       hwData->readIndex = ((hwData->readIndex+1) % hwData->addrCount);
-      if ( handleCount > 1000 ) break;
    }
 
    // Process transmit software queue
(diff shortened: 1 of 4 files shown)


ruck314 (approved):

slacrherbst:
Here are some comparative rates using a single channel firmware prbs generator into a software receiver. The new driver rate and bandwidth are increased in irq mode. The poll mode results in a performance hit.

Old Driver
-------------

Size | Rate | BW
-- | -- | --
16384 | 3.05E+04 | 4.00E+09
1024 | 4.04E+05 | 3.31E+09
128 | 7.30E+05 | 7.47E+08
64 | 7.30E+05 | 3.74E+08
 
New Driver, IRQ
---------------------

Size | Rate | BW
-- | -- | --
16384 | 3.05E+04 | 4.00E+09
1024 | 4.86E+05 | 3.94E+09
128 | 7.66E+05 | 7.88E+08
64 | 7.47E+05 | 3.97E+08

New Driver, Poll
---------------------

Size | Rate | BW
-- | -- | --
 16384 | 3.05E+04 | 4.00E+09
 1024 | 4.86E+05 | 3.98E+09
 128 | 6.67E+05 | 6.83E+08
 64 | 6.72E+05 | 3.44E+08

PR #101, opened 2023-02-14
Add generic load script with success checking, fix for kernels > 5.18

slacrherbst:
The first change is the addition of a load_datadev.sh script which can be used to load the driver with configuration values as well as check the success of the load and determine if hardware was properly found.

This is mean to address ESROGUE-596.

This PR also includes a fix which addresses issues with kernels > 5.18.

DmaWrite has a few updates plus there is a new test.cpp script.


ruck314:
@slacrherbst Have you've regression tested in hardware?

bengineerd (approved):

PR #104, opened 2023-03-10
adding petalinux/aximemorymap to support PS I/O peripherals (SPI, I2C, etc)

ruck314:
### Description
- This is required because on the RFSoC 4x2 board the LMK/LMX is only connected on the PS (not the PL)
- In other RFSoC boards, the LMK was connected to both PS and PL.  This is why it was not an issue in the past.

### DIFF
--- a/petalinux/aximemorymap/files/aximemorymap.c
+++ b/petalinux/aximemorymap/files/aximemorymap.c
@@ -35,8 +35,11 @@
 // Module Name
 #define MOD_NAME "axi_memory_map"
 
-unsigned long cfgMinAddr = 0x400000000; // Edit this to match your AXI port address configurations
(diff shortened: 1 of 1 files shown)


ruck314:
Now that I can access SPI0 on the ZynqMP device, I can create a SPI bridge AXI-Lite proxy to convert AXI-Lite transactions into this SPI transactions.
https://www.xilinx.com/htmldocs/registers/ug1087/mod___spi.html
image

ruck314:
Confirmed the out-of-bound range checking works:
```
[  852.288212] axi_memory_map Map_Find: Invalid address 0xfe040004
[  852.288212]  PS Allowed range 0xff000000 - 0xffffffff
[  852.288212]  PL Allowed range 0x400000000 - 0x4ffffffff
```

slacrherbst (approved):

PR #113, opened 2024-02-14
Fix bad mapping call for new kernels

slacrherbst:
Fix bad mapping call for new kernels.

This line was in error meaning it was calling the deprecated mapping function instead dma_mmap_coherent.


ruck314 (approved):
I have confirmed this does work and fixes the bug for both kernel 5.15.0 and kernel 6.5.0 hardware

PR #118, opened 2024-02-22
bug fixes and updates to data_dev_top.c

ruck314:
### Description
- [code clean up using 'Kernel-doc' style](https://github.com/slaclab/aes-stream-drivers/commit/0eb9342a4bcfde5726dbaa38a11f10fe455a313f)
- [use readl() instead of ioread32()](https://github.com/slaclab/aes-stream-drivers/pull/118/commits/256bc402f514328b3c60a4d43cbe297b2e3d0415)
  - These changes replace the direct I/O access functions with the recommended memory-mapped I/O functions, improving portability and security within the kernel module.
- [Error Handling for pci_enable_device()](https://github.com/slaclab/aes-stream-drivers/pull/118/commits/16fb81121ae2a2009614841285cc95293b76bfb6)
  - Before accessing the device's resources, you must call pci_enable_device. This function can fail, so check its return value
- [use snprintf() w/ error checking instead of sprintf()](https://github.com/slaclab/aes-stream-drivers/pull/118/commits/208080cb9b6cbdce12b1990db2fff3676b6098bc)
  - snprintf() call ensures that the output is properly truncated if it exceeds the size of the destination buffer, thus preventing buffer overflow vulnerabilities
- [dma_set_mask_and_coherent is deprecated in favor of separate calls to dma_set_mask and dma_set_coherent_mask separately](https://github.com/slaclab/aes-stream-drivers/pull/118/commits/f193f36adbd4b8b4602271a31de2f57ebed04160)
  - using separate calls provides greater clarity, flexibility, and compatibility with future kernel versions, while potentially offering performance benefits and easing debugging efforts
- [polishing DataDev_Probe() code comments](https://github.com/slaclab/aes-stream-drivers/pull/118/commits/51d4b2373c561e78a4f72fa52435e95cc86db0c1)
- [adding cfgDevName feature to use the PCI device bus number for unique device naming](https://github.com/slaclab/aes-stream-drivers/pull/118/commits/3b2aabcf9a4b357c9a59e3b3324a253cf97d6867)
  - Helpful when multiple PCIe cards are installed in the same server

### DIFF
--- a/data_dev/driver/src/data_dev_top.c
+++ b/data_dev/driver/src/data_dev_top.c
@@ -247,13 +247,20 @@ int DataDev_Probe(struct pci_dev *pcidev, const struct pci_device_id *dev_id) {
    AxiVersion_SetUserReset(dev->base + AVER_OFF,false);
 
    // 128bit desc, = 64-bit address map
-   if ( (readl(dev->reg) & 0x10000) != 0) {
+   if ((readl(dev->reg) & 0x10000) != 0) {
       // Get the AXI Address width (in units of bits)
-      axiWidth = (readl(dev->reg+0x34) >> 8) & 0xFF;
-      if ( !dma_set_mask_and_coherent(dev->device, DMA_BIT_MASK(axiWidth)) ) {
-         dev_info(dev->device,"Init: Using %d-bit DMA mask.\n",axiWidth);
+      axiWidth = (readl(dev->reg + 0x34) >> 8) & 0xFF;
+
+      if (!dma_set_mask(dev->device, DMA_BIT_MASK(axiWidth))) {
+         dev_info(dev->device, "Init: Using %d-bit DMA mask.\n", axiWidth);
+
+         if (!dma_set_coherent_mask(dev->device, DMA_BIT_MASK(axiWidth))) {
+            dev_info(dev->device, "Init: Using %d-bit coherent DMA mask.\n", axiWidth);
+         } else {
+            dev_warn(dev->device, "Init: Failed to set coherent DMA mask.\n");
+         }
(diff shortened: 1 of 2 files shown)


ruck314:
Confirmed this worked using DMA loopback firmware on the following Linux kernels:
- 5.15.0-94-generic
- 6.5.0-18-generic

data_dev/driver/src/data_dev_top.c
```diff
@@ -67,234 +66,347 @@ MODULE_DEVICE_TABLE(pci, DataDev_Ids);
...
+
+      // Attempt to set DMA and coherent DMA masks based on AXI width
+      if (!dma_set_mask(dev->device, DMA_BIT_MASK(axiWidth))) {
+         dev_info(dev->device, "Init: Using %d-bit DMA mask.\n", axiWidth);
+
+         if (!dma_set_coherent_mask(dev->device, DMA_BIT_MASK(axiWidth))) {
+            dev_info(dev->device, "Init: Using %d-bit coherent DMA mask.\n", axiWidth);
+         } else {
+            dev_warn(dev->device, "Init: Failed to set coherent DMA mask.\n");
+         }
```
JJL772, line 271: Are failures here not considered fatal? Kernel docs say we should not attempt to use DMA until `dma_set_mask` returns success.

JJL772 (approved):
Looks good besides one comment/question.

data_dev/driver/src/data_dev_top.c
ruck314: I'll defer that question to @slacrherbst, who wrote the driver

PR #131, opened 2024-05-02
Fix bug in AxisG2_Clear when 128-bit descriptors are enabled

JJL772:
Fix bug in AxisG2_Clear when 128-bit descriptors are enabled

### Description

`dma_free_coherent` was being called with a size argument assuming descriptors were 64-bits. This doesn't match the call to `dma_alloc_coherent` which properly computes DMA size depending on the descriptor size.

This pretty much just copies that logic over to `AxisG2_Clear`.

### JIRA

https://jira.slac.stanford.edu/browse/ESROGUE-666


ruck314:
@JJL772 Does this fix the issue of [ESROGUE-666](https://jira.slac.stanford.edu/browse/ESROGUE-666)?

JJL772:
Yes

ruck314 (approved):

PR #135, opened 2024-05-06
Move common driver code into `common/driver/data_dev_common.c`

JJL772:
### Description

This PR pulls common code out of DataGpu/DataDev and puts it in `common/driver/data_dev_common.c`. When working on #133, I had to apply the error handling changes in both drivers. This PR remedies that.

### Details

- `DataDev_Common_Probe` takes a callback `int(*InitDevCallback)(struct DmaDevice* dev)` which is called when the `DmaDevice`'s parameters need to be initted (i.e. `cfgTxCount`, `cfgSize`, etc.). The module parameter code is the only place where `data_gpu` and `data_dev` diverged in their respective probe functions, hence this change.
- `struct pci_driver* gPciDriver`, `const char* gModName` and `struct hardware_functions* gHardwareFuncs` need to be defined in each module's top, as they are referenced from the common code.
- `module_init`/`module_exit` only work with functions defined in the translation unit where `module_init`/`module_exit` lie. So, I opted to create module-specific functions that proxy into common driver code, even if no additional behavior is required. For example, `DataGpu_Init` simply calls `DataDev_Common_Init` and nothing else.

Draft for now. Mostly looking for feedback before I proceed with testing.


ruck314:
From Jeremy
```
I left it as a draft to gather feedback before continuing to spend time on it. Still needs some work on my part and I believe there are some conflicts by now.```

PR #136, opened 2024-05-23
update to Dma_Ioctl()

ruck314:
### Description
- Adding DMA_Get_GITV to expose DMA GIT version string to userspace


ruck314:
Confirmed this works:
image

common/driver/dma_common.c
```diff
@@ -1025,6 +1025,14 @@ ssize_t Dma_Ioctl(struct file *filp, uint32_t cmd, unsigned long arg) {
          }
          break;
 
+      // Get GIT Version
+      case DMA_Get_GITV:
+         if (copy_to_user((char *)arg, GITV, sizeof(GITV))) {
```
JJL772, line 1030: I assume `GITV` here is a string macro? Perhaps `sizeof(GITV)` should be replaced with something like `min(sizeof(GITV), MAX_GITV_LENGTH)` to ensure there is a minimum safe buffer size.

ruck314: https://github.com/slaclab/aes-stream-drivers/blob/main/data_dev/driver/Makefile#L45
@JJL772 `GITV` is an env var string in the Makefile.  Do you still think this min. buffer size check is required?

JJL772: Sorry for the late response. I suppose it's okay to leave it without bounds. My only concern is if someone creates an excessively long tag that unexpectedly exceeds buffer size in the calling code. Normally this should not be a problem though, these tags are pretty short.

ruck314: [0471c4f](https://github.com/slaclab/aes-stream-drivers/pull/136/commits/0471c4ffe459ea1ed7571a00d26a8e9f838b057b) addresses your concerns

JJL772: Thanks!

JJL772 (approved):

PR #138, opened 2024-05-28
Update comp_and_load_drivers.sh

ruck314:
### Description
- Add check_sudo() function to the beginning of the script


JJL772 (approved):
That was fast!

PR #151, opened 2024-07-09
overhauling the comp_and_load_drivers.sh script

ruck314:
### Description
- Adding CC arg to datagpu Makefile
- Added instructions to README.md about disabling systemctl services that automatically load the driver without the p2p configurations
- fixed the broken kernel gcc finder in the bash script


rclaus2 (approved):
As near as I can tell, these changes look okay to me.  Both the script and the Makefile appeared to work correctly on rdsrv419.

PR #160, opened 2024-07-31
Move 'DMA mapping failed' print to the failure handling case of the c…

rclaus2:
…onditional

Printing of 'DMA mapping failed' was being done when dma_mapping_error() returned false instead of when it returned true.

### Description

The change was to move the 'DMA mapping failed' print to the case when dma_mapping_error() returned true.  This avoids this print appearing for each DMA buffer registered in dmesg output and in /var/log/messages.


JJL772 (approved):
Thanks for catching this!

PR #162, opened 2024-10-16
Update DmaDriver.h

ruck314:
### Description
- Applying the linter fix

### DIFF
--- a/common/app_lib/PrbsData.cpp
+++ b/common/app_lib/PrbsData.cpp
@@ -22,6 +22,8 @@
 #include <stdlib.h>
 #include <stdio.h>
 
+#include <cstdio>
+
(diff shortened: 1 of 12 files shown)


JJL772 (approved):
Linter seems unhappy in a few files due to `#include`'s

ruck314:
@JJL772 Yeah, still working through them

common/app_lib/PrbsData.cpp
```diff
@@ -22,6 +22,8 @@
 #include <stdlib.h>
 #include <stdio.h>
```
bengineerd, line 23: I think we need to include either <stdio.h> or <stdio> but not both.

JJL772:
Oops, did not realize this was a draft!

PR #177, opened 2025-01-27
Corrected buffer offset address to align it with the firmware in AxiPcieGpuAsyncControl.vhd

mmishra9:
### Description
Added the correct offset to be subtracted from the 'mapSize' to align the 'mapSize' with 'maxSize' in firmware: https://github.com/slaclab/axi-pcie-core/blob/main/protocol/gpuAsync/rtl/AxiPcieGpuAsyncControl.vhd#L299

### Details
Because the DATA_BYTE_C (32 bytes) is added to the descriptor address here: https://github.com/slaclab/axi-pcie-core/blob/main/protocol/gpuAsync/rtl/AxiPcieGpuAsyncControl.vhd#L307,
we should subtract the same offset when we set the 'maxSize': https://github.com/slaclab/axi-pcie-core/blob/main/protocol/gpuAsync/rtl/AxiPcieGpuAsyncControl.vhd#L299

### DIFF
--- a/common/driver/gpu_async.c
+++ b/common/driver/gpu_async.c
@@ -171,8 +171,8 @@ int32_t Gpu_AddNvidia(struct DmaDevice *dev, uint64_t arg) {
          // Update buffer count and write DMA addresses to device
          if (buffer->write) {
             writel(buffer->dmaMapping->dma_addresses[0] & 0xFFFFFFFF, data->base+0x100+data->writeBuffers.count*16);
-            writel((buffer->dmaMapping->dma_addresses[0] >> 32) & 0xFFFFFFFF, data->base+0x104+data->writeBuffers.count*16); 
-            writel(mapSize - BUFFER_OFFSET, data->base+0x108+data->writeBuffers.count*16); // subtracted BUFFER_OFFSET to align the mapSize with maxSize in firmware
+            writel((buffer->dmaMapping->dma_addresses[0] >> 32) & 0xFFFFFFFF, data->base+0x104+data->writeBuffers.count*16);
+            writel(mapSize - BUFFER_OFFSET, data->base+0x108+data->writeBuffers.count*16);  // Subtracted BUFFER_OFFSET to align the mapSize with maxSize in firmware
             data->writeBuffers.count++;
          } else {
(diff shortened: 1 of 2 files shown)


ruck314:
@mmishra9 
```
Run find . -name '*.h' -o -name '*.cpp' -o -name '*.c' | xargs cpplint
./common/driver/gpu_async.c:174:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
./common/driver/gpu_async.c:175:  At least two spaces is best between code and comments  [whitespace/comments] [2]
./data_gpu/driver/src/gpu_async.c:174:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
./data_gpu/driver/src/gpu_async.c:175:  At least two spaces is best between code and comments  [whitespace/comments] [2]
```

JJL772 (approved):
Thanks! I do wonder if this will cause unexpected behavior though (i.e. you give it a 32768 byte buffer, but only 32512 bytes are usable.)

Perhaps we should return the real buffer size somehow? Or expose the number of additional bytes that need to be added to buffer allocations?

JJL772 (changes requested):
@mmishra9  Looking at this again, `BUFFER_OFFSET_SHIFT` is 8, so `BUFFER_OFFSET` ends up being 255. This seems rather large and doesn't match the `+32` that's in firmware.

I'm not sure I really understand why this is the case; is `BUFFER_OFFSET` in bits instead of bytes?

mmishra9:
Yes, **BUFFER_OFFSET** is in bits and **+32** that's in the firmware is in bytes, so 256 bits = 32 bytes.

mmishra9:
After talking to @JJL772, I updated the **BUFFER_OFFSET_SHIFT** to 5; it should be in byte, i.e., 32 bytes.

JJL772:
`BUFFER_OFFSET` is defined to `(BUFFER_OFFSET_SIZE - 1)`, which works out to be 31. Do we want the `- 1` in there still?

mmishra9:
No, it should be 32. I updated it. Thanks.

JJL772 (approved):
Thanks for fixing this @mmishra9!

PR #166, opened 2024-11-12
/proc/datagpu improvements and GpuAsyncRegs.h header

JJL772:
### Description

Expands on the /proc/datagpu_N interface. Exposes additional internal state from GpuAsyncCore, as well as the gpu async core offset.

Also adds a register definitions header to help avoid magic numbers. Looking for feedback on this file. I was trying to avoid writing 3 preprocessor defines for each register (offset, bit offset, mask), as it's extremely verbose.

PR #179, opened 2025-02-07
Fix signed overflow in /proc/datadev_N "GPU State" section

JJL772:
### Description

Ric reports that RX frame count goes negative sometimes. This is a mistake I made in #166

### DIFF
--- a/common/driver/gpu_async.c
+++ b/common/driver/gpu_async.c
@@ -314,7 +314,7 @@ int32_t Gpu_SetWriteEn(struct DmaDevice *dev, uint64_t arg) {
  * @dev: Device to read from
  */
 void Gpu_Show(struct seq_file *s, struct DmaDevice *dev) {
-   int i;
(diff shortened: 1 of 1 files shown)


common/driver/gpu_async.c
```diff
@@ -329,10 +329,10 @@ void Gpu_Show(struct seq_file *s, struct DmaDevice *dev) {
    seq_printf(s, "           Write Enable : %d\n", writeEnable);
    seq_printf(s, "      Read Buffer Count : %d\n", readBuffCnt);
```
ruck314, line 330: @JJL772 While you are at it, can you update the integer type of other non-negative values as well?  E.g. readBuffCnt is also never negative as well and probably should be `%u\n`

JJL772: Done

ruck314 (approved):

PR #192, opened 2025-03-15
Handle non-page aligned GPU memory addresses

JJL772:
### Description

CUDA doesn't provide a way to perform arbitrary alignment memory allocations, it only guarantees us 256-byte alignment with cudaMalloc. We need to make sure to handle this case in the driver.

For what it's worth, cudaMalloc(0x10000) does spit out 64k aligned device addresses, but, as far as I can tell, this isn't guaranteed.

This PR addresses the case where the device address is not aligned to a 64k boundary, and/or when the buffer size is not a multiple of 64k. 

In the case where size % 64k != 0, the FPGA may overwrite another buffer's memory if the buffers are tightly packed.
For example:
Buff0    - 0xB0000 - sz: 4096
Buff1    - 0xB1000 - sz: 4096
...
If the FPGA is given mapSize=64k (which is what will currently happen with the contiguous memory check), it'll write unexpected junk to buff1 too. cudaMalloc tends to return sequential blocks like that.

It may be advisable to enforce a 64k-aligned size constraint as well.

### DIFF
--- a/common/driver/gpu_async.c
+++ b/common/driver/gpu_async.c
@@ -110,7 +110,7 @@ int32_t Gpu_Command(struct DmaDevice *dev, uint32_t cmd, uint64_t arg) {
 int32_t Gpu_AddNvidia(struct DmaDevice *dev, uint64_t arg) {
    int32_t ret;
    uint32_t x;
-   u64     virt_start;
+   u64     virt_start, virt_offset, dma_address;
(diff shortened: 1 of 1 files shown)


mmishra9 (approved):
Adding the offset to the dma address makes sense. It should fix the issue I observed.

PR #194, opened 2025-03-24
Constrain buffer sizes to multiple of 64kb and add cfgTimeout parameter

JJL772:
### Description

* Added cfgTimeout parameter to the driver
* Added 64kb buffer size align requirement to workaround issues not fully fixed by #192. This is more of a workaround than anything else.
* Handle timeout field of the descriptor data


JJL772:
Tested. @mmishra9 I'm still getting timeouts even with a timeout of 10 million though

JJL772:
Seems to work with dmaRate and the CUDA-based test_dma.

ruck314 (approved):

PR #202, opened 2025-06-20
Report error if the PCI device is missing an IRQ

JJL772:
### Description

Report an error and return failure in `DataDev_Probe` if the PCIE device is missing an IRQ.

@ruck314 is it ever valid to have a device without an associated IRQ? If so, this can be made into a warning.

### JIRA

https://jira.slac.stanford.edu/browse/ESROGUE-713

### DIFF
--- a/common/driver/data_dev_top.c
+++ b/common/driver/data_dev_top.c
@@ -251,6 +251,13 @@ int DataDev_Probe(struct pci_dev *pcidev, const struct pci_device_id *dev_id) {
    // Assign the IRQ number from the pci_dev structure
    dev->irq = pcidev->irq;
 
+   // Check that we actually have an IRQ
+   if (dev->irq == 0) {
+      pr_err("%s: No IRQ associated with PCI device\n", MOD_NAME);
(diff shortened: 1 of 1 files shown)


ruck314:
Confirmed that this works
```bash
[   17.418140] datadev: Init
[   17.418167] datadev 0000:01:00.0: enabling device (0000 -> 0002)
[   17.418172] (NULL device *): Init: Mapping Register space 0xf3000000 with size 0x1000000.
[   17.418280] (NULL device *): Init: Mapped to 0xffff99e485000000.
[   17.418282] datadev: No IRQ associated with PCI device
[   17.418671] datadev: probe of 0000:01:00.0 failed with error -22
[   17.418698] datadev: Init: failure detected in init. Unregistering driver.
```
@JJL772 While it is possible to run the driver without a IRQ, it is a requirement that the FPGA firmware supports interrupt.  This new error checking makes sure that the IP core for the PCIe on the FPGA is configured properly.

ruck314 (approved):

PR #244, opened 2026-02-07
Avoid allocations during runtime

JJL772:
### Description

Pre-allocates scratch buffers so we don't have to keep calling `kmalloc` during `dmaReadIndex` and such.

Introduces the following limitations:
* Max number of indices that can be returned at once is cfgRxCount + cfgTxCount
* Max number of buffers that can be read or written at once is cfgRxCount + cfgTxCount

If those conditions are violated, the ioctl returns with error.

I am not sure if the DAQ is trying to read/write/return more than its allotted amount.

### DIFF
--- a/common/driver/dma_common.c
+++ b/common/driver/dma_common.c
@@ -477,10 +477,16 @@ int Dma_Open(struct inode *inode, struct file *filp) {
    }
 
    // Allocate scratch data buffers
-   desc->readDataScratch = (struct DmaReadData*)kzalloc(sizeof(struct DmaReadData) * READ_DATA_SCRATCH_CNT, GFP_KERNEL);
-   desc->indexScratch = (uint32_t*)kzalloc(sizeof(uint32_t) * INDEX_SCRATCH_CNT, GFP_KERNEL);
+   desc->readScratchCount = dev->cfgRxCount;
+   desc->readDataScratch = (struct DmaReadData*)kzalloc(sizeof(struct DmaReadData) * desc->readScratchCount, GFP_KERNEL);
 
-   if (!desc->readDataScratch || !desc->indexScratch) {
+   desc->indexScratchCount = MAX(dev->cfgRxCount, dev->cfgTxCount);
+   desc->indexScratch = (uint32_t*)kzalloc(sizeof(uint32_t) * desc->indexScratchCount, GFP_KERNEL);
+
+   desc->buffListScratchCount = MAX(dev->cfgRxCount, dev->cfgTxCount);
+   desc->buffListScratch = (struct DmaBuffer**)kzalloc(sizeof(struct DmaBuffer*) * desc->buffListScratchCount, GFP_KERNEL);
+
+   if (!desc->readDataScratch || !desc->indexScratch || !desc->buffListScratch) {
       dev_err(dev->device, "Open: kzalloc for scratch area(s) failed\n");
       kfree(desc);
       return -ENOMEM;
@@ -610,14 +616,16 @@ static ssize_t Dma_Read_Internal(struct file *filp, __user char *buffer, size_t 
    struct DmaDevice *dev = desc->dev;
 
    // Sanity check!
-   if (rCnt > READ_DATA_SCRATCH_CNT) {
-      dev_err(dev->device, "Dma_Read_Internal: Programming error: rCnt > READ_DATA_SCRATCH_CNT (%d)\n",
-         READ_DATA_SCRATCH_CNT);
+   if (rCnt > desc->readScratchCount) {
+      dev_err(dev->device, "Dma_Read_Internal: Programming error: rCnt > desc->readScratchCount (%d)\n",
+         desc->readScratchCount);
       return -1;
    }
 
    //struct DmaBuffer* buff[NUM_BUFFERS_AT_ONCE] = {0};
-   struct DmaReadData rd[READ_DATA_SCRATCH_CNT] = {0};
+   //struct DmaReadData rd[READ_DATA_SCRATCH_CNT] = {0};
+
+   struct DmaReadData* rd = desc->readDataScratch;
 
    // Copy the read structure from user space
    if ((ret = copy_from_user(rd, buffer, rCnt * sizeof(struct DmaReadData)))) {
@@ -732,7 +740,7 @@ ssize_t Dma_Read(struct file *filp, __user char *buffer, size_t count, loff_t *f
 
    // Read at most READ_DATA_SCRATCH_CNT per iteration. Avoids allocations in the driver.
    for (; rem > 0;) {
-      size_t toRead = MIN(rem, READ_DATA_SCRATCH_CNT);
+      size_t toRead = MIN(rem, desc->readScratchCount);
       ret = Dma_Read_Internal(filp, buffer, toRead);
 
       // If failed, this will be a partial read. This should only happen in exceptionally rare
@@ -887,17 +895,17 @@ static ssize_t Dma_Ret_Indexes(struct file *filp, uint32_t cmd, __user void* arg
       return -1;
    }
 
-   for (int i = 0; i < cnt; i += INDEX_SCRATCH_CNT) {
-      uint32_t indexes[INDEX_SCRATCH_CNT] = {0};
-      struct DmaBuffer* buffList[INDEX_SCRATCH_CNT] = {0};
+   mutex_lock(&desc->mutex);
 
+   for (uint32_t i = 0; i < cnt; i += desc->indexScratchCount) {
       // Limit the number to free per iteration
-      const size_t toFree = MIN(INDEX_SCRATCH_CNT, cnt - i);
+      const size_t toFree = MIN(desc->indexScratchCount, cnt - i);
 
       // This should really never happen because we already checked with access_ok
-      if (copy_from_user(indexes, arg, (toFree * sizeof(uint32_t)))) {
+      if (copy_from_user(desc->indexScratch, arg, (toFree * sizeof(uint32_t)))) {
          dev_warn(dev->device, "Ret_Indexes: copy_from_user failed. buf=%p, bytes=%ld\n",
             arg, toFree * sizeof(uint32_t));
+         mutex_unlock(&desc->mutex);
          return -1;
       }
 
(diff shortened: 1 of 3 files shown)


JJL772:
At this point, I'm pretty confident that these changes will not cause issues with the DAQ

common/driver/dma_common.c
```diff
@@ -476,7 +476,25 @@ int Dma_Open(struct inode *inode, struct file *filp) {
...
-   memset(desc, 0, sizeof(struct DmaDesc));
+   // Allocate scratch data buffers
+   desc->readScratchCount = dev->cfgRxCount + dev->cfgTxCount;
+   desc->readDataScratch = (struct DmaReadData*)kzalloc(sizeof(struct DmaReadData) * desc->readScratchCount, GFP_KERNEL);
+
+   desc->indexScratchCount = dev->cfgRxCount + dev->cfgTxCount;
+   desc->indexScratch = (uint32_t*)kzalloc(sizeof(uint32_t) * desc->indexScratchCount, GFP_KERNEL);
+
+   desc->buffListScratchCount = dev->cfgRxCount + dev->cfgTxCount;
+   desc->buffListScratch = (struct DmaBuffer**)kzalloc(sizeof(struct DmaBuffer*) * desc->buffListScratchCount, GFP_KERNEL);
```
ruck314, line 493: **Must-fix:** these three allocations are leaked on close. `Dma_Release` (lines 520-599, unchanged in this PR) calls `dmaQueueFree(&desc->q)` and `kfree(desc)` but never frees `readDataScratch`, `indexScratch`, or `buffListScratch`. Every FD open/close cycle now leaks `(cfgRxCount+cfgTxCount) * (sizeof(DmaReadData)+sizeof(uint32_t)+sizeof(DmaBuffer*))` bytes — easy to demonstrate with `for i in $(seq 1 10000); do exec 9</dev/datadev_0; exec 9<&-; done` and `slabtop`.

Please add the three `kfree`s in `Dma_Release` before `kfree(desc);`.

common/driver/dma_common.c
```diff
@@ -476,7 +476,25 @@ int Dma_Open(struct inode *inode, struct file *filp) {
...
+
+   if (!desc->readDataScratch || !desc->indexScratch || !desc->buffListScratch) {
+      dev_err(dev->device, "Open: kzalloc for scratch area(s) failed\n");
+      kfree(desc);
```
ruck314, line 500: **Must-fix:** partial-allocation leak. If e.g. the first two `kzalloc`s succeed but the third fails, this path frees only `desc` and leaks the two successful allocations. `kfree(NULL)` is a no-op, so unconditionally `kfree`ing all three scratch pointers here is fine.

common/driver/dma_common.h
```diff
@@ -184,6 +189,23 @@ struct DmaDesc {
...
+
+   // Scratch data buffers for ioctl and read operations
+   uint32_t* indexScratch;
+   struct DmaReadData* readDataScratch;
+   struct DmaBuffer** buffListScratch;
+
+   // Number of elements in the read scratch buffer. Corresponds to cfgRxCount
+   uint32_t readScratchCount;
+
+   // Number of elements in the index scratch buffer. max(cfgTxCount, cfgRxCount)
```
ruck314: **Should-fix:** comments are stale. All three counts are allocated as `cfgRxCount + cfgTxCount` in `Dma_Open` (dma_common.c:480, 483, 486) — not `max(cfgTxCount, cfgRxCount)` (this comment and the next), and not just `cfgRxCount` (`readScratchCount` comment above). Please update the prose to match the code (or, if `max()` was actually the intent, change the code — sum is a safe upper bound, but the comments and code need to agree).

common/driver/dma_common.c
```diff
@@ -594,62 +612,77 @@ int Dma_Release(struct inode *inode, struct file *filp) {
...
-         (ulong)(rCnt * sizeof(struct DmaReadData)));
-      return -ENOMEM;
+   rem = rCnt;
+
+   // Check that we can read that many buffers
+   if (rCnt > desc->readScratchCount || rCnt > desc->buffListScratchCount) {
+      dev_warn(dev->device, "Read: attempted to read too many buffers. rCnt=%zu > max=%d\n",
+         rCnt, MIN(desc->readScratchCount, desc->buffListScratchCount));
+      mutex_unlock(&desc->mutex);
+      return -1;
```
ruck314: **Should-fix:** raw `-1` is `-EPERM` to userspace, which is misleading. The PR already uses `-EFAULT` (line 656) and `-EINVAL` (line 902) elsewhere in the same paths — please use proper `-E*` codes here and at the analogous returns at line 666 (copy_from_user fail in `Dma_Read`), line 868 (over-cap in `Dma_Ret_Indexes`), and line 878 (copy_from_user fail in `Dma_Ret_Indexes`). Likely `-EINVAL` for the cap violations and `-EFAULT` for the copy failures.

common/driver/dma_common.c
```diff
@@ -800,6 +833,87 @@ ssize_t Dma_Write(struct file *filp, __user const char *buffer, size_t count, lo
...
+   uint32_t bCnt = 0;
+
+   desc = (struct DmaDesc *)filp->private_data;
+   dev  = desc->dev;
+
+   // Ensure the buffer is readable
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5, 0, 0)
+   if (!access_ok(arg, cnt * sizeof(uint32_t), VERIFY_READ)) {
+#else
+   if (!access_ok(arg, cnt * sizeof(uint32_t))) {
```
ruck314, line 880: **Should-fix:** `access_ok` is called with the unvalidated user-supplied `cnt` (up to `0xFFFF` → 256 KB) before the `cnt > desc->indexScratchCount` bound check at line 865. Reorder so the bound check runs first — that keeps a pathological caller from handing a giant size to `access_ok`.

common/driver/dma_common.c
```diff
@@ -594,62 +612,77 @@ int Dma_Release(struct inode *inode, struct file *filp) {
...
+   size_t rem = 0;
+   ssize_t bCnt = 0;
+   size_t copied = 0;
+   ssize_t x = 0;
+   __user void *dp = NULL;
+   struct DmaDesc *desc = (struct DmaDesc *)filp->private_data;
+   struct DmaDevice *dev = desc->dev;
+   struct DmaReadData *rd = NULL;
+
+   mutex_lock(&desc->mutex);
```
ruck314: **Nice-to-have:** the mutex is held across `copy_from_user` (line 662), `copy_to_user` (lines 705, 723), and per-buffer hardware ops — all of which can sleep/page-fault. Same pattern in `Dma_Ret_Indexes`. For single-threaded-per-FD DAQ use this is harmless, but it does serialize otherwise-independent ops on the same FD. Either narrow the lock to just protect the scratch arrays themselves, or leave a brief comment explaining the chosen scope.

common/driver/dma_common.h
```diff
@@ -53,6 +53,11 @@ struct DmaDesc;
 typedef uint32_t __poll_t;
 #endif
 
+#undef MIN
+#define MIN(_x, _y) ((_x) < (_y) ? (_x) : (_y))
```
ruck314: **Nice-to-have:** the kernel already provides type-checked `min()`/`max()` in `<linux/minmax.h>`. These macros are unsafe (no parens around the macro arguments → operator-precedence surprises) and only `MIN` is used in one place at dma_common.c:642. Suggest dropping the macros and using `min()` at the single call site.

common/driver/dma_common.c
```diff
@@ -685,15 +718,15 @@ ssize_t Dma_Read(struct file *filp, __user char *buffer, size_t count, loff_t *f
...
-               ret, buffer, &rd);
+   if ((copied = copy_to_user(buffer, rd, rCnt * sizeof(struct DmaReadData)))) {
+      dev_warn(dev->device, "Read: failed to copy struct to user space ret=%zu, user=%p kern=%p\n",
+               copied, buffer, &rd);
    }
-   kfree(rd);
-   return bCnt;
+
+   mutex_unlock(&desc->mutex);
+   return x;
```
ruck314, line 741: **Nice-to-have:** semantically equivalent to `return bCnt;` (the loop runs `x` from 0 to `bCnt` with no `break`), but `return bCnt;` reads more clearly and matches the doxygen comment above ("The number of read structures on success").

common/driver/dma_common.c
```diff
@@ -800,6 +833,87 @@ ssize_t Dma_Write(struct file *filp, __user const char *buffer, size_t count, lo
...
+ * Dma_Ret_Indexes - Return indexes to the hardware
+ * @filp: File pointer
+ * @cmd: IOCTL command to execute
+ * @arg: user destination buffer
+ */
+static ssize_t Dma_Ret_Indexes(struct file *filp, uint32_t cmd, __user void* arg) {
+   struct DmaDesc   * desc;
+   struct DmaDevice * dev;
+   struct DmaBuffer * buff;
+   uint32_t cnt = (cmd >> 16) & 0xFFFF;
```
ruck314, line 858: **Nice-to-have:** the previous `DMA_Ret_Index` ioctl had an `if (cnt == 0) return 0;` early-exit that's no longer here. Not a bug — the loop just doesn't iterate — but a no-op `access_ok` + `copy_from_user` for size 0 is wasted work. Cheap to restore.

ruck314 (changes requested):
Thanks for tackling this — moving the kmalloc/kfree off the DMA hot path is a clear win, and the conversion of `Dma_Read`/`DMA_Ret_Index` is mostly mechanical and clean. A few things to address before merge.

**Must-fix:** the three scratch buffers allocated in `Dma_Open` are leaked on close (`Dma_Release` never frees them), and there's a partial-allocation leak on `kzalloc` failure.
**Should-fix:** stale header comments, raw `-1` returns where `-EINVAL`/`-EFAULT` belong, and `access_ok` ordering in `Dma_Ret_Indexes`.
**Nice-to-have:** the kernel already provides type-checked `min()`/`max()`, plus a couple of small readability/efficiency nits.

I cross-checked the other open PRs (#264, #265, #267, #268) — none touch the scratch-buffer paths, so these comments aren't redundant with anything pending. Userspace ABI is preserved (`dmaRetIndex`/`dmaRetIndexes`/`dmaReadIndex` in `include/DmaDriver.h` unchanged) and `userHas == desc` ownership is still enforced, so once the leaks are fixed and the error codes tightened this should be ready.

Inline comments below.

common/driver/dma_common.c
JJL772: We need to ensure the scratch buffers are untouched for the duration of the work (from copy_from_user through the retRxBuffer). To avoid page faults under the mutex, we can pin the pages first (and unpin when finished), however I don't know what the overhead of this is.

JJL772: Pinning might not be worth it. `get_user_pages_fast` and `get_user_pages` is typically suggested for short-lived pins, but it needs to be paired with `put_page` for releasing, but `put_page` must be called per-page. Thus, if we're pinning 256 pages, we have to call `put_page` 256 times at the end of the syscall.

`get_user_pages_fast` itself is pretty much free if the pages are resident already, it's just the unpinning part that sucks.

ruck314 (approved):

ruck314:
@JJL772 LGTM. Merge when you are ready

PR #258, opened 2026-04-08
only remove old driver if its present

tomeichlersmith:
I noticed this while doing a few power cycles. The rc.local job fails at start-up because `rmmod` fails. A simple guard on if the driver is already loaded before we try to remove it is helpful so that the `load` script can be used by both the `rc.local` and by a user during driver development.


JJL772 (approved):
LGTM, thanks!

PR #260, opened 2026-04-14
Refactor DKMS stuff

JJL772:
### Description
- Builds DKMS via the Makefile instead of only in CI.
- Also adds DKMS configuration for the GPU variant of the module.


ruck314 (approved):
@JJL772 LGTM. Merge when you are ready

PR #267, opened 2026-04-29
Fix bring-up / teardown bugs in data_dev, aximemorymap, axistreamdma

ruck314:
## Summary

Reviewed and fixed bring-up (probe / `_Init`) and teardown (remove / `_Exit`) sequences for three kernel drivers and the shared `dma_common` helpers they call:

- `data_dev/driver/src/data_dev_top.c` (PCIe `datadev` driver — symlink to `common/driver/data_dev_top.c`)
- `Yocto/recipes-kernel/aximemorymap/files/aximemorymap.c`
- `Yocto/recipes-kernel/axistreamdma/files/axistreamdma.c`

While in there, also fixed a pre-existing version-guard bug that was preventing builds against kernel 6.2 / 6.3.

### Resource leaks on probe failure paths
- **`data_dev_top.c`**: `DataDev_Probe` failures after `Dma_MapReg` succeeded (IRQ-missing check, `dma_set_mask` / `dma_set_coherent_mask` failures) only called `pci_disable_device` and leaked the `iomap` + `mem_region` claim. Added `err_unmap` label that calls `Dma_UnmapReg` before disabling the PCI device.
- **`axistreamdma.c`**: `Rce_Probe` failures after `irq_of_parse_and_map` and `Dma_MapReg` used bare `return -1` and leaked the virtual IRQ mapping plus the ioremap + mem_region claim. Added `err_post_irq_map` / `err_post_mapreg` / `err_dma_init` labels that unwind in reverse order, including `irq_dispose_mapping` and a `memset` to clear the `gDmaDevices` slot.

### Teardown bugs
- **`aximemorymap.c`**: `Map_Exit` was missing `cdev_del(&dev.charDev)`, leaving the character device live across module unload. Teardown order also did not match the reverse of `Map_Init` (`class_destroy` ran after `unregister_chrdev_region`). Reordered to: `cdev_del` -> `device_destroy` -> `class_destroy` -> `unregister_chrdev_region` -> free maps.
- **`axistreamdma.c`**: `Rce_Remove` never called `irq_dispose_mapping` on the virtual IRQ created by `irq_of_parse_and_map` — every probe -> remove cycle leaked an IRQ mapping. (`Dma_Clean` is intentionally untouched so the PCI path through `data_dev_top`, which uses `pdev->irq` directly, is unaffected.)

### API-correctness bugs in helper code
- **`dma_common.c` / `aximemorymap.c`**: `class_create()` and `device_create()` return `ERR_PTR` on failure, not `NULL`. The existing `== NULL` checks let real failures through to a kernel oops on `gCl->devnode = ...`. Switched to `IS_ERR_OR_NULL()`. On `class_create` failure, also reset `gCl = NULL` so the shared `(gDmaDevCount == 0 && gCl != NULL)` cleanup logic stays safe.
- **`dma_common.c` / `aximemorymap.c`**: `cdev_add()` returns `-errno` on failure (`-EBUSY`, `-ENOMEM`, …), but the checks were `== -1` and only caught `-EPERM`. Switched to `< 0`.
- **`dma_common.c`**: `Dma_UnmapReg` was non-idempotent (no NULL check, no clear of `dev->base`) and `Dma_MapReg` left `dev->base` pointing at unmapped memory after a `request_mem_region` failure. Made `Dma_UnmapReg` idempotent and `Dma_MapReg` clear `dev->base` on error so multi-path cleanup is safe.

### Kernel version guards
- **`axistreamdma.c` / `axistreamdma.h`**: the platform-driver `.remove` callback returns `int` prior to kernel 6.11 and `void` from 6.11 onward. The current `void Rce_Remove(...)` only matched 6.11+, so the driver did not compile on the supported kernel range (3.10 - 6.10). Version-guarded both the prototype and the definition.
- **`dma_common.c` / `dma_common.h` / `aximemorymap.c`**: the `class.devnode()` callback became `const struct device *` in upstream commit `ff62b8e6588f` ("driver core: make struct class.devnode() take a const *") in v6.2-rc1, not v6.4. The version guard for `Dma_DevNode` / `Map_DevNode` was conflated with the separate `class_create()` `THIS_MODULE`-removal change in v6.4 (commit `1aaba11da9aa`), causing a function-pointer type mismatch when building against kernels 6.2 and 6.3. Lowered only the devnode guard to `KERNEL_VERSION(6, 2, 0)` (the `class_create()` guard at 6.4 stays — that change really is in 6.4). Kept the `RHEL_RELEASE_VERSION(9, 4)` fallback for both, matching how RHEL 9.4 backported the upstream changes.

## Commits

Five atomic commits, smallest blast radius first:

1. `dma_common: harden init/cleanup error paths and helper APIs`
2. `data_dev_top: unmap registers on probe failures after Dma_MapReg`
3. `aximemorymap: fix Map_Exit teardown order and init error checks`
4. `axistreamdma: clean up probe/remove resource paths and guard remove signature`
5. `dma_common, aximemorymap: lower class.devnode const guard from 6.4 to 6.2`

## Test plan

- [x] `cpplint` (matching CI invocation) clean on all modified files
- [x] CI whitespace / tab checks clean
- [x] Local `make driver` against all installed kernels: 5.15.0-91, 5.15.0-139, 5.19.0-50, 6.2.0-39, 6.5.0-45, 6.8.0-59, 6.8.0-107, 6.8.0-110 — all build clean. (Previously 6.2.0-39 failed with `-Werror=incompatible-pointer-types`.)
- [x] Live smoke test on a real `/dev/datadev_0`: two `rmmod datadev` + `insmod datadev.ko` cycles on running 6.8.0-110 kernel — module reloaded cleanly each time, `/dev/datadev_0` re-created, IRQ 162 reacquired, no warnings / oops / leak messages in dmesg, `/proc/iomem` shows a single `f3000000-f3ffffff : datadev_0` claim with no duplication.
- [x] CI build matrix (Ubuntu 22.04 / 24.04 / Rocky 9 / Debian experimental / CentOS 7) — to verify on PR
- [x] Yocto / platform-driver path — needs target hardware to validate `axistreamdma` rmmod/insmod cycle and confirm IRQ-mapping leak fix

## Notes for review

- All fixes preserve existing behavior on the success path; only error / teardown paths change.
- No public API change. The only internal-helper signature change is `Dma_UnmapReg` becoming idempotent (no callers needed updating).
- `Dma_Clean` is left unchanged — the PCI path through `data_dev_top` uses `pdev->irq` directly and never calls `irq_of_parse_and_map`, so disposing the mapping there would be wrong. Only `axistreamdma`'s `Rce_Remove` does it explicitly.
- Sister drivers `rce_stream/driver/src/rce_top.c` and `rce_hp_buffers/driver/src/rce_top.c` likely share the same IRQ-mapping leak and inverse `.remove` signature problem (`int` -> breaks on 6.11+). Out of scope for this PR.


common/driver/dma_common.c
```diff
@@ -142,7 +142,9 @@ struct class *gCl;
  *
  * Returns NULL always.
  */
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 4, 0) || (defined(RHEL_RELEASE_CODE) && RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(9, 4))
+// class.devnode() became const in v6.2 (upstream commit ff62b8e6588f),
+// backported by RHEL 9.4. Distinct from the class_create() change in v6.4.
```
JJL772: These comments should be less verbose and match the doxygen style

common/driver/dma_common.h
```diff
@@ -227,7 +227,11 @@ extern struct class * gCl;
 extern struct file_operations DmaFunctions;
 
 // Function prototypes
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 4, 0) || (defined(RHEL_RELEASE_CODE) && RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(9, 4))
+// class.devnode() took a non-const struct device* until upstream commit
+// ff62b8e6588f ("driver core: make struct class.devnode() take a const *")
+// in v6.2-rc1; RHEL 9.4 backported it. Note: this is a different change
+// from the class_create() signature change in v6.4 below.
```
JJL772: Comment is too verbose

Yocto/recipes-kernel/axistreamdma/files/axistreamdma.c
```diff
@@ -302,6 +304,18 @@ int Rce_Probe(struct platform_device *pdev) {
...
+   irq_dispose_mapping(dev->irq);
+   memset(dev, 0, sizeof(*dev));
+   return -1;
+
+err_post_mapreg:
+   Dma_UnmapReg(dev);
+err_post_irq_map:
+   irq_dispose_mapping(dev->irq);
+   memset(dev, 0, sizeof(*dev));
+   return -1;
```
JJL772, line 313: Duplicate label. Same content as `err_dma_init`

Yocto/recipes-kernel/axistreamdma/files/axistreamdma.c
```diff
@@ -315,10 +329,17 @@ int Rce_Probe(struct platform_device *pdev) {
  *
  * @pdev: Platform device structure representing the DMA device.
  */
+// The platform_driver .remove callback returns int prior to kernel 6.11
+// and void from 6.11 onward; mirror that with a version-guarded signature.
```
JJL772: Unnecessary comment

Yocto/recipes-kernel/axistreamdma/files/axistreamdma.h
```diff
@@ -49,7 +50,14 @@ int Rce_Probe(struct platform_device *pdev);
  * This function performs the cleanup and removal of the device driver
  * associated with a given platform device. It is called by the Linux
  * kernel during the device removal process.
+ *
+ * The platform_driver .remove callback signature changed from int to
+ * void in kernel 6.11; the prototype tracks that.
```
JJL772: Unnecessary comment

JJL772 (changes requested):

common/driver/dma_common.c
ruck314: Agreed — dropped the comment. The function already has a doxygen header above it, and the `#if` guard is self-documenting; the historical detail (upstream commit ff62b8e6588f, RHEL 9.4 backport) lives in the commit log for the change that introduced the guard. Fixed in ed8e913.

common/driver/dma_common.h
ruck314: Agreed — dropped the 4-line block from the header; the `#if` guard alone is sufficient. Fixed in ed8e913.

Yocto/recipes-kernel/axistreamdma/files/axistreamdma.c
ruck314: Good catch — `err_dma_init` had the same body as `err_post_irq_map`. Removed the duplicate label and `Dma_Init` failure now `goto err_post_irq_map` directly (skipping `err_post_mapreg` because `Dma_Init` has already unwound `Dma_MapReg` on its failure path). Fixed in ed8e913.

Yocto/recipes-kernel/axistreamdma/files/axistreamdma.c
ruck314: Agreed — the `#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 11, 0)` is self-documenting. Removed. Fixed in ed8e913.

Yocto/recipes-kernel/axistreamdma/files/axistreamdma.h
ruck314: Agreed — removed the paragraph from the doxygen block. The `#if` guard on the prototype communicates the same thing. Fixed in ed8e913.

JJL772 (approved):

PR #270, opened 2026-04-30
Driver bring-up / teardown audit + docs touch-up

ruck314:
## Summary

Audit of the kernel driver bring-up and teardown paths in `common/driver/`, `data_dev/driver/src/`, `Yocto/recipes-kernel/aximemorymap/files/`, `Yocto/recipes-kernel/axistreamdma/files/`, and `rce_hp_buffers/driver/src/`. Each commit fixes one issue and is meant to be readable on its own. A small docs change rides along (one stale how-to dropped, one architecture note added to match the new `init` contract).

## Fixes (commit-by-commit)

1. **`dma_buffer: harden dmaQueueFree against NULL queue->queue`** — `dmaQueueInit` can fail mid-way and leave `queue->queue == NULL` while `queue->subCount` is already set. Walking the array in that state NULL-derefs. Guard the walk and trailing kfree.
2. **`axis_gen2: propagate AxisG2_Init allocation failures via int return`** — `AxisG2_Init` had five unchecked allocations (hwData, two ring buffers via `dma_alloc_coherent` or `kzalloc`, and `buffList`); under OOM the next deref crashed. Change `hardware_functions::init` to return `int`, NULL-check each allocation, free what was already taken via a goto-chain, and leave `dev->hwData == NULL` on failure. `AxisG1_Init` and `RceHp_Init` become `int` and trivially return 0. `Dma_Init` now honors the return code via a new `cleanup_rx_buffers` label that skips `->clear()` (which would deref the half-built hwData).
3. **`data_dev_top: unmap registers when Dma_Init's early paths fail`** — `Dma_Init`'s early failure paths (chrdev/class/device/proc) skip `cleanup_dma_mapreg` and so leak the register region the caller mapped. Route the failure goto through `err_unmap`. `Dma_UnmapReg` is idempotent, so a late-path Dma_Init failure that already unmapped is unaffected.
4. **`data_dev_top: free GPU utilData on remove and probe failure`** — `Gpu_Init` kzallocs a `GpuData` and stashes it in `dev->utilData`. `gpu_async.c` has no teardown, so every datadev unload (and every probe failure after `Gpu_Init` ran) leaked it. Free in `DataDev_Remove` before `Dma_Clean`, and on the `err_unmap` failure path. Both gated on `DATA_GPU` so non-GPU builds are unchanged. The fix lives in `data_dev_top.c` (rather than `gpu_async.c`) because the latter is gated by the CI's "gpu_async.c unchanged from main" check.
5. **`axistreamdma: unmap registers when Dma_Init's early paths fail`** — same class of bug as #3 in the platform driver; route through `err_post_mapreg`. Replace the previous comment that incorrectly claimed `Dma_Init` always unwinds its own register mapping.
6. **`aximemorymap: add THIS_MODULE owner to MapFunctions`** — `MapFunctions` was missing `.owner = THIS_MODULE`. Userspace can hold an open fd while `rmmod` unloads the driver, then call ioctl on the stale fd and use-after-free into freed module text.
7. **`aximemorymap: pass devName via %s to device_create`** — `device_create`'s name argument is a printf format string. `dev.devName` is a fixed literal today, but the call should not be sensitive to that.
8. **`axis_gen2: use GFP_KERNEL for buffList in AxisG2_Init`** — `AxisG2_Init` runs in process context (probe path), so the `buffList` `kzalloc` should be `GFP_KERNEL` rather than `GFP_KERNEL | GFP_DMA32`. The DMA32 zone is irrelevant for an internal pointer array and just needlessly constrains the allocator.

## Docs

- **`docs(how-to): drop github-pages-setup guide`** — one-time setup instructions that no longer reflect how the docs deploy works; removed from `docs/how-to/` and the toctree.
- **`docs(architecture): note int return contract for hardware_functions::init`** — reflects #2 above. Updates the `hardware_functions` code block in `docs/explanation/architecture.rst` and adds a paragraph explaining the success/failure semantics and why the common layer skips `clear()` on failure.

The Doxygen-driven API reference (`docs/reference/*.rst`) auto-renders from the public userspace headers (`DmaDriver.h`, `AxisDriver.h`, `GpuAsync*.h`), which are unchanged in this PR; nothing to update there.

## Test plan

- [x] CI Phase 1 (lint, doxygen, trailing-whitespace, gpu_async.c unchanged gate) passes
- [x] CI Phase 2 CPU emulator load/unload cycle on each distro completes without dmesg warnings
- [x] CI Phase 3 GPU emulator path passes — the `data_dev_top: free GPU utilData ...` change is the most behavior-affecting one for that path
- [x] Out-of-tree probe-failure smoke (force a kzalloc/ioremap failure via fault injection) — optional, but the most direct way to exercise the new error paths in `AxisG2_Init`, `Dma_Init`, and `DataDev_Probe`
- [x] Local: `make` cleanly builds `data_dev/driver` and `aximemorymap` against a 6.x kernel
- [x] Local: `cd docs && doxygen Doxyfile && make html` builds without new warnings

## Out of scope (noted, not fixed)

While reviewing, I noticed `rce_stream/driver/src/rce_top.c` has the same `Dma_Init`-leak pattern as the platform/PCI drivers, plus a `gDmaDevCount++` placed before any failure check, leaving the count incorrect on every probe failure. It is the legacy V3 RCE driver and was not in the dirs originally requested for this audit; happy to address in a follow-up.

### DIFF
--- a/common/driver/axis_gen1.c
+++ b/common/driver/axis_gen1.c
@@ -189,14 +189,17 @@ int AxisG1_Init(struct DmaDevice *dev) {
    iowrite32(0x1, &(reg->rxEnable));
    iowrite32(0x1, &(reg->txEnable));
 
-   // Push RX buffers to hardware
+   // Push RX buffers to hardware. A per-buffer map failure here means
+   // the RX ring would come up partially primed -- a broken driver
+   // state -- so fail init and let Dma_Init unwind via cleanup_rx_buffers.
+   // AxisG1_Init owns no allocations, so a bare error return is sufficient.
    for (x=dev->rxBuffers.baseIdx; x < (dev->rxBuffers.baseIdx + dev->rxBuffers.count); x++) {
       buff = dmaGetBufferList(&(dev->rxBuffers), x);
       if ( dmaBufferToHw(buff) < 0 ) {
-         dev_warn(dev->device, "Init: Failed to map dma buffer.\n");
-      } else {
-          iowrite32(buff->buffHandle, &(reg->rxFree));
+         dev_err(dev->device, "Init: Failed to map dma buffer.\n");
+         return -EIO;
       }
+      iowrite32(buff->buffHandle, &(reg->rxFree));
    }
 
    // Set cache mode
--- a/common/driver/axis_gen1.c
+++ b/common/driver/axis_gen1.c
@@ -197,7 +197,7 @@ int AxisG1_Init(struct DmaDevice *dev) {
       buff = dmaGetBufferList(&(dev->rxBuffers), x);
       if ( dmaBufferToHw(buff) < 0 ) {
          dev_err(dev->device, "Init: Failed to map dma buffer.\n");
-         return -EIO;
+         return -ENOMEM;
       }
       iowrite32(buff->buffHandle, &(reg->rxFree));
    }
--- a/common/driver/axis_gen1.c
+++ b/common/driver/axis_gen1.c
@@ -171,7 +171,7 @@ irqreturn_t AxisG1_Irq(int irq, void *dev_id) {
 
 
 // Init card in top level Probe
-void AxisG1_Init(struct DmaDevice *dev) {
+int AxisG1_Init(struct DmaDevice *dev) {
    uint32_t x;
 
    struct DmaBuffer  *buff;
@@ -209,6 +209,7 @@ void AxisG1_Init(struct DmaDevice *dev) {
    // Set dest mask
    memset(dev->destMask, 0xFF, DMA_MASK_SIZE);
    dev_info(dev->device, "Init: Found Version 1 Device.\n");
+   return 0;
 }
 
 // Enable or disable IRQs in the hardware
--- a/common/driver/axis_gen2.c
+++ b/common/driver/axis_gen2.c
@@ -539,13 +539,19 @@ int AxisG2_Init(struct DmaDevice *dev) {
    if ( dev->version >= 3 ) writel(dev->cfgIrqHold, &(reg->irqHoldOff));
    if ( dev->version >= 5 ) writel(dev->cfgTimeout, &(reg->timeout));
 
-   // Push RX buffers to hardware and map
+   // Push RX buffers to hardware and map. A per-buffer map failure means
+   // the RX ring would come up partially primed -- a broken driver state
+   // -- so fail init and unwind through the cleanup chain.
    for (x=dev->rxBuffers.baseIdx; x < (dev->rxBuffers.baseIdx + dev->rxBuffers.count); x++) {
       buff = dmaGetBufferList(&(dev->rxBuffers), x);
 
       // Map failure
       if ( dmaBufferToHw(buff) < 0 ) {
-          dev_warn(dev->device, "Init: Failed to map dma buffer.\n");
+         dev_err(dev->device, "Init: Failed to map dma buffer.\n");
+         if (dev->cfgMode & AXIS2_RING_ACP)
+            goto err_free_writeaddr_acp;
+         else
+            goto err_free_writeaddr_dma;
 
       // Add to software queue, if enabled and hardware is full
       } else if ( hwData->desc128En && (hwData->hwWrBuffCnt >= (hwData->addrCount-1)) ) {
@@ -572,7 +578,14 @@ int AxisG2_Init(struct DmaDevice *dev) {
 
    // Allocation failure cleanup. dev->hwData is left NULL so
    // AxisG2_Clear (not invoked here) is never called with a
-   // half-built hwData.
+   // half-built hwData. RX buffers already pushed to hardware
+   // are reclaimed by Dma_Init via cleanup_rx_buffers.
+err_free_writeaddr_dma:
+   dma_free_coherent(dev->device, size, hwData->writeAddr, hwData->writeHandle);
+   goto err_free_readaddr_dma;
+err_free_writeaddr_acp:
+   kfree(hwData->writeAddr);
+   goto err_free_readaddr_acp;
 err_free_readaddr_dma:
    dma_free_coherent(dev->device, size, hwData->readAddr, hwData->readHandle);
    goto err_free_bufflist;
--- a/common/driver/axis_gen2.c
+++ b/common/driver/axis_gen2.c
@@ -407,7 +407,7 @@ irqreturn_t AxisG2_Irq(int irq, void *dev_id) {
  * This includes setting up DMA buffers, configuring hardware registers, and initializing
  * software queues for efficient DMA transfers.
  */
-void AxisG2_Init(struct DmaDevice *dev) {
+int AxisG2_Init(struct DmaDevice *dev) {
    uint32_t x;
    uint32_t size;
 
@@ -423,6 +423,10 @@ void AxisG2_Init(struct DmaDevice *dev) {
 
    // Allocate and initialize hardware data structure
    hwData = (struct AxisG2Data *)kzalloc(sizeof(struct AxisG2Data), GFP_KERNEL);
+   if (hwData == NULL) {
+      dev_err(dev->device, "Init: Failed to allocate AxisG2Data.\n");
+      return -ENOMEM;
+   }
    dev->hwData = hwData;
    hwData->dev = dev;
 
@@ -436,11 +440,28 @@ void AxisG2_Init(struct DmaDevice *dev) {
    hwData->hwWrBuffCnt = 0;
    hwData->hwRdBuffCnt = 0;
 
-   // Initialize software queues if in 128-bit descriptor mode
+   // Initialize software queues if in 128-bit descriptor mode.
+   // dmaQueueInit returns 0 on success-with-zero-elements *and*
+   // on failure, so test queue->queue (NULL on failure) for an
+   // unambiguous result. The cleanup goto for each step targets
+   // the matching err_free_<queue> label so the partially-built
+   // queue is freed via dmaQueueFree.
    if ( hwData->desc128En ) {
       dmaQueueInit(&hwData->wrQueue, dev->rxBuffers.count);
+      if (hwData->wrQueue.queue == NULL) {
+         dev_err(dev->device, "Init: Failed to init wrQueue.\n");
+         goto err_free_wrqueue;
+      }
       dmaQueueInit(&hwData->rdQueue, dev->txBuffers.count + dev->rxBuffers.count);
-      hwData->buffList = (struct DmaBuffer **)kzalloc(BUFF_LIST_SIZE * sizeof(struct DmaBuffer *), GFP_ATOMIC);
+      if (hwData->rdQueue.queue == NULL) {
+         dev_err(dev->device, "Init: Failed to init rdQueue.\n");
+         goto err_free_rdqueue;
+      }
+      hwData->buffList = (struct DmaBuffer **)kzalloc(BUFF_LIST_SIZE * sizeof(struct DmaBuffer *), GFP_KERNEL);
+      if (hwData->buffList == NULL) {
+         dev_err(dev->device, "Init: Failed to allocate buffList.\n");
+         goto err_free_rdqueue;
+      }
    }
 
    // Calculate and set the addressable space based on register settings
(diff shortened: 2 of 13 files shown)


common/driver/axis_gen2.c
```diff
@@ -438,9 +442,20 @@ void AxisG2_Init(struct DmaDevice *dev) {
...
-      dmaQueueInit(&hwData->wrQueue, dev->rxBuffers.count);
-      dmaQueueInit(&hwData->rdQueue, dev->txBuffers.count + dev->rxBuffers.count);
+      if (dmaQueueInit(&hwData->wrQueue, dev->rxBuffers.count) == 0 && dev->rxBuffers.count > 0) {
+         dev_err(dev->device, "Init: Failed to init wrQueue.\n");
+         goto err_free_hwdata;
+      }
+      if (dmaQueueInit(&hwData->rdQueue, dev->txBuffers.count + dev->rxBuffers.count) == 0 &&
+          (dev->txBuffers.count + dev->rxBuffers.count) > 0) {
+         dev_err(dev->device, "Init: Failed to init rdQueue.\n");
+         goto err_free_wrqueue;
```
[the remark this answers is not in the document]
ruck314: Valid — but routing through `err_free_rdqueue` directly would have caused a *double-free* in `dmaQueueFree`, because the deeper bug was in `dmaQueueInit`: its `cleanup_sub_queue` label freed the sub-queue pointers but neither freed `queue->queue` nor NULLed the entries it freed. Fixed both:

- `785d73d` — `dmaQueueInit` cleanup now `kfree`s `queue->queue`, sets it to NULL, and NULLs each freed sub-queue entry. So a failed init leaves the queue fully clean.
- `0ac1853` — failure check switched to `queue->queue == NULL` (unambiguous after the fix above). Goto target stays at `err_free_wrqueue` so the conventional "free everything before the failed step" cleanup chain is preserved; with the cleanup fix, calling `dmaQueueFree` on a failed queue is now a safe no-op.

common/driver/axis_gen2.c
```diff
@@ -438,9 +442,20 @@ void AxisG2_Init(struct DmaDevice *dev) {
...
    // Initialize software queues if in 128-bit descriptor mode
    if ( hwData->desc128En ) {
-      dmaQueueInit(&hwData->wrQueue, dev->rxBuffers.count);
-      dmaQueueInit(&hwData->rdQueue, dev->txBuffers.count + dev->rxBuffers.count);
+      if (dmaQueueInit(&hwData->wrQueue, dev->rxBuffers.count) == 0 && dev->rxBuffers.count > 0) {
+         dev_err(dev->device, "Init: Failed to init wrQueue.\n");
+         goto err_free_hwdata;
+      }
+      if (dmaQueueInit(&hwData->rdQueue, dev->txBuffers.count + dev->rxBuffers.count) == 0 &&
+          (dev->txBuffers.count + dev->rxBuffers.count) > 0) {
```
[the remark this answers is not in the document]
ruck314: Valid — applied in `0ac1853`. The check is now `hwData->wrQueue.queue == NULL` (and likewise for `rdQueue`), which is unambiguous regardless of count. This relies on `785d73d` to also NULL `queue->queue` on every `dmaQueueInit` failure path; together they catch the count==0 + alloc-fail edge case.

common/driver/dma_common.c
```diff
@@ -339,8 +339,13 @@ int Dma_Init(struct DmaDevice *dev) {
    if ( dev->cfgRxCount > 0 && res == 0 )
       goto cleanup_dma_queue;
 
-   // Call card specific init
-   dev->hwFunc->init(dev);
+   // Call card specific init. On failure the init function
+   // owns its own cleanup; we must not call ->clear() because
+   // dev->hwData may be NULL or partially built.
+   if (dev->hwFunc->init(dev) < 0) {
+      dev_err(dev->device, "Init: Card-specific init failed.\n");
```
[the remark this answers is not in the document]
ruck314: Valid — applied in `0ac1853`. Captured into `res` and logged via `%zd`. `Dma_Init` itself still collapses to a single `return -1`; propagating real errnos through `Dma_Init` is a larger refactor (separate from the bring-up/teardown audit), but at least the actual init errno now appears in dmesg.

common/driver/data_dev_top.c
```diff
@@ -315,10 +316,14 @@ int DataDev_Probe(struct pci_dev *pcidev, const struct pci_device_id *dev_id) {
...
    }
 
-   // Initialize common DMA functionalities
+   // Initialize common DMA functionalities. Dma_Init's early
+   // failure paths (chrdev/class/device/proc) do not unmap the
+   // registers we mapped above, so route through err_unmap.
+   // Dma_UnmapReg is idempotent, so a double-unmap from a late
+   // Dma_Init failure path is harmless.
    if (Dma_Init(dev) < 0) {
       probeReturn = -ENOMEM;      // Indicate memory allocation error
```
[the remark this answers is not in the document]
ruck314: Valid — applied in `2ad74a7`. `probeReturn` now takes `Dma_Init`'s actual return code (`probeReturn = Dma_Init(dev); if (probeReturn < 0) goto err_unmap;`). Today `Dma_Init` always returns `-1`, but once it propagates real errnos this code stays correct.

common/driver/dma_common.h
```diff
@@ -221,7 +221,7 @@ struct DmaDesc {
  */
 struct hardware_functions {
    irqreturn_t (*irq)(int irq, void *dev_id);
-   void        (*init)(struct DmaDevice *dev);
+   int         (*init)(struct DmaDevice *dev);
```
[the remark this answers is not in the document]
ruck314: Valid and critical catch — `RceHp_Init` was indeed still `void`; fixed in `0ac1853`. Updated `rce_hp_buffers/driver/src/rce_hp.{c,h}` so `RceHp_Init` returns `int` and trivially returns `0`. The audit was originally scoped to `common/driver/`, `data_dev/`, and the two Yocto trees — `rce_hp_buffers` slipped through. Thanks for catching this one.

common/driver/axis_gen2.c
```diff
@@ -438,9 +442,20 @@ void AxisG2_Init(struct DmaDevice *dev) {
 
    // Initialize software queues if in 128-bit descriptor mode
    if ( hwData->desc128En ) {
-      dmaQueueInit(&hwData->wrQueue, dev->rxBuffers.count);
-      dmaQueueInit(&hwData->rdQueue, dev->txBuffers.count + dev->rxBuffers.count);
+      if (dmaQueueInit(&hwData->wrQueue, dev->rxBuffers.count) == 0 && dev->rxBuffers.count > 0) {
+         dev_err(dev->device, "Init: Failed to init wrQueue.\n");
+         goto err_free_hwdata;
```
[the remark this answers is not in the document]
ruck314: Same root cause as the `rdQueue` review on this file: `dmaQueueInit`'s `cleanup_sub_queue` was leaking `queue->queue`, so simply rerouting the goto would have caused a double-free in `dmaQueueFree`. Fixed in:

- `785d73d` — `dmaQueueInit` now fully cleans up on every failure path (`kfree(queue->queue); queue->queue = NULL;` and NULLs the freed sub-queue entries).
- `0ac1853` — failure detection in `AxisG2_Init` now reads `hwData->wrQueue.queue == NULL` directly. Goto stays at `err_free_hwdata` since with the cleanup fix the failed wrQueue holds nothing to free.

common/driver/axis_gen2.c
```diff
@@ -436,11 +440,28 @@ void AxisG2_Init(struct DmaDevice *dev) {
...
+      if (hwData->wrQueue.queue == NULL) {
+         dev_err(dev->device, "Init: Failed to init wrQueue.\n");
+         goto err_free_wrqueue;
+      }
       dmaQueueInit(&hwData->rdQueue, dev->txBuffers.count + dev->rxBuffers.count);
+      if (hwData->rdQueue.queue == NULL) {
+         dev_err(dev->device, "Init: Failed to init rdQueue.\n");
+         goto err_free_rdqueue;
+      }
       hwData->buffList = (struct DmaBuffer **)kzalloc(BUFF_LIST_SIZE * sizeof(struct DmaBuffer *), GFP_ATOMIC);
```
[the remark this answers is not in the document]
ruck314: Valid — applied in `0cec07a`. Switched `buffList` to `GFP_KERNEL` to match the rest of `AxisG2_Init` (which is called from probe context via `Dma_Init` and is fully sleepable). The `GFP_ATOMIC` was a leftover from `db49745` with no atomic-context justification — every other allocation in this function (`hwData`, ring buffers, etc.) already uses `GFP_KERNEL`.

common/driver/axis_gen1.c
JJL772: Since this now returns an error code (0 in all paths), should this be treated as an error?

common/driver/axis_gen2.c
JJL772: Should this constitute an error?

common/driver/axis_gen1.c
ruck314: Valid catch — once `init` returns `int` specifically to surface bring-up failures, silently warning on a per-buffer `dmaBufferToHw` failure during the RX-prime loop is exactly the inconsistency this PR was meant to remove. A partially primed RX ring is a broken driver state, and `dmaBufferToHw`'s contract (`Returns 0 on success, or -1 on error`) leaves room for real failure modes even though the current implementation always returns 0.

Fixed in `b0625a4`: `dev_warn` → `dev_err` + `return -EIO`. `AxisG1_Init` owns no allocations, so a bare error return is sufficient — `Dma_Init` routes through `cleanup_rx_buffers` and reclaims the RX list.

common/driver/axis_gen2.c
ruck314: Valid and fixed in b0625a4 — same reasoning as the `axis_gen1` comment: once `init` returns `int` to surface bring-up failures, a per-buffer map failure during the RX-prime loop should fail init rather than warn-and-continue. Partial RX ring = broken driver state.

Fixed in `b0625a4`. `AxisG2_Init` is the harder case because the failure point is after `writeAddr` is allocated, so the cleanup chain needed two new labels:

```
err_free_writeaddr_dma:  dma_free_coherent(writeAddr) → goto err_free_readaddr_dma
err_free_writeaddr_acp:  kfree(writeAddr)             → goto err_free_readaddr_acp
```

The new failure routes through the matching ACP/DMA label and unwinds via the existing chain. RX buffers already pushed to hardware are reclaimed by `Dma_Init`'s `cleanup_rx_buffers`.

common/driver/axis_gen1.c
```diff
@@ -189,14 +189,17 @@ int AxisG1_Init(struct DmaDevice *dev) {
...
+   // state -- so fail init and let Dma_Init unwind via cleanup_rx_buffers.
+   // AxisG1_Init owns no allocations, so a bare error return is sufficient.
    for (x=dev->rxBuffers.baseIdx; x < (dev->rxBuffers.baseIdx + dev->rxBuffers.count); x++) {
       buff = dmaGetBufferList(&(dev->rxBuffers), x);
       if ( dmaBufferToHw(buff) < 0 ) {
-         dev_warn(dev->device, "Init: Failed to map dma buffer.\n");
-      } else {
-          iowrite32(buff->buffHandle, &(reg->rxFree));
+         dev_err(dev->device, "Init: Failed to map dma buffer.\n");
+         return -EIO;
```
JJL772: This should return `-ENOMEM`. AxisG2_Init returns `-ENOMEM` in this same situation and the return values should match between the two.

common/driver/axis_gen1.c
ruck314: Addressed in b0625a4 — AxisG1_Init now returns an error (and AxisG2_Init gotos through its cleanup chain) when dmaBufferToHw fails.

common/driver/axis_gen2.c
ruck314: Addressed in b0625a4 — AxisG2_Init now gotos through err_free_writeaddr_{dma,acp} on dmaBufferToHw failure, returning -ENOMEM after unwinding.

common/driver/axis_gen1.c
ruck314: Good catch — fixed in 9d6e884. Now returns -ENOMEM to match AxisG2_Init.

JJL772 (approved):

PR #273, opened 2026-05-08
Add gpuEnableRx/Tx, avoid setting internal regs in rdmaTest

JJL772:
### Description
- These are internal registers and should not be set by userspace in most cases.

### DIFF
--- a/include/GpuAsyncUser.h
+++ b/include/GpuAsyncUser.h
@@ -104,11 +104,13 @@ public:
 
    /**
     * @brief Sets the number of write buffers.
+    * @param val Number of write buffers.
     * @note gpuAddNvidiaMemory sets this for you.
     * @note This is an INTERNAL register, you should not use this unless you know what you're doing!
     */
    inline void setWriteCount(uint32_t val) {
-      writeRegV1V4(GpuAsyncReg_WriteCountV1, GpuAsyncReg_WriteCountV4, val-1);
+      // Count is zero inclusive, but we've made it cleaner for callers of this function.
+      writeRegV1V4(GpuAsyncReg_WriteCountV1, GpuAsyncReg_WriteCountV4, val ? val-1 : 0);
    }
 
    inline uint32_t writeEnable() const {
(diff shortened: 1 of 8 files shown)


include/GpuAsyncUser.h
```diff
@@ -102,14 +102,23 @@ class GpuAsyncCoreRegs {
...
    }
 
+   /**
+    * @brief Sets the number of write buffers.
+    * @note gpuAddNvidiaMemory sets this for you.
+    * @note This is an INTERNAL register, you should not use this unless you know what you're doing!
+    */
    inline void setWriteCount(uint32_t val) {
-      writeRegV1V4(GpuAsyncReg_WriteCountV1, GpuAsyncReg_WriteCountV4, val);
+      writeRegV1V4(GpuAsyncReg_WriteCountV1, GpuAsyncReg_WriteCountV4, val-1);
```
ruck314: @JJL772 Maybe add a note in the comments above that writeCount is zero inclusive

include/GpuAsyncUser.h
```diff
@@ -118,14 +127,23 @@ class GpuAsyncCoreRegs {
...
    }
 
+   /**
+    * @brief Sets the number of read buffers.
+    * @note gpuAddNvidiaMemory sets this for you.
+    * @note This is an INTERNAL register, you should not use this unless you know what you're doing!
+    */
    inline void setReadCount(uint32_t val) {
-      writeRegV1V4(GpuAsyncReg_ReadCountV1, GpuAsyncReg_ReadCountV4, val);
+      writeRegV1V4(GpuAsyncReg_ReadCountV1, GpuAsyncReg_ReadCountV4, val-1);
```
ruck314: @JJL772 Maybe add a note in the comments above that writeCount is zero inclusive

include/GpuAsyncUser.h
JJL772: Done

include/GpuAsyncUser.h
```diff
@@ -118,14 +129,25 @@ class GpuAsyncCoreRegs {
       return readRegV1V4(GpuAsyncReg_ReadCountV1, GpuAsyncReg_ReadCountV4);
```
ruck314: @JJL772 Should this get() function return a +1 of the memory read value to be normalized to the `-1` in the set() function?

include/GpuAsyncUser.h
```diff
@@ -102,14 +102,25 @@ class GpuAsyncCoreRegs {
       return readRegV1V4(GpuAsyncReg_WriteCountV1, GpuAsyncReg_WriteCountV4);
```
ruck314: @JJL772 Should this get() function return a +1 of the memory read value to be normalized to the `-1` in the set() function?

JJL772: Good catch .That seems like a good idea...

ruck314 (approved):
@JJL772 LGTM. Merge when ready

PR #274, opened 2026-05-09
axis_gen2: fix TX continue-bit drop + drop unused label/dead code

ruck314:
## Summary

Three small, separable commits:

1. **`axis_gen2: drop unused err_free_hwdata label`** — silences the `-Wunused-label` warning emitted on every build of `AxisG2_Init`. No goto ever targeted `err_free_hwdata:`; it was reached only by fallthrough from `err_free_wrqueue:`. Removing the label preserves the cleanup statements (and their fallthrough order) — pure dead-syntax removal.

2. **`axis_gen2: preserve continue bit in AxisG2_WriteTx`** — *real bug fix.* Lines 178–179 both used `=` instead of the second using `|=`, so the **continue** bit (bit[3]) was silently overwritten by the **timeout** bit (bit[4]) before either reached the descriptor. Lines 180–181 already use `|=`. Net effect on `main`/`pre-release` today: every TX submission ships with the continue flag forced to 0 regardless of `buff->flags`. One-character fix (`=` → `|=`).

3. **`common/driver: drop dead var and redundant kfree NULL guards`** — pure cleanup. Removes an unused `size_t rem` in `Dma_Read` (compiler-confirmed `-Wunused-but-set-variable`), and three redundant `if (x != NULL) kfree(x);` guards in `dmaAllocBuffers`'s cleanup chain. `kfree(NULL)` is documented kernel-safe; verified `dmaFreeBuffersList` does not touch the freed arrays so no double-free risk either way.

Net diff: `+4 / −8` across `axis_gen2.c`, `dma_common.c`, `dma_buffer.c`.

## Reviewer please also look at

While auditing for the same class of issue I noticed a parallel sloppy spot I left **out of scope** so this PR stays narrow:

- **`common/driver/dma_buffer.c:255-258`** in `dmaFreeBuffers` — same `if (list->indexed != NULL) kfree(list->indexed); if (list->sorted != NULL) kfree(list->sorted);` pattern as the cleanup path I just trimmed in `dmaAllocBuffers`. Drop the guards here too for parity, or push back if you'd rather keep the defensive style and have me restore the guards in commit 3 instead. Either is fine; I just wanted both cleanup functions to read the same way and didn't want to silently expand scope.

## Test plan

- [x] `make clean && make` in `data_dev/driver/` — clean build, only the pre-existing toolchain "compiler differs" notice remains; the `-Wunused-label err_free_hwdata` warning is gone
- [x] CI trailing-whitespace / tab gate — clean on all three edited files
- [x] `cpplint` on the three edited files — clean
- [x] Full CI pipeline (CPU + GPU matrix across 5 distros, DKMS smoke) — will run on push
- [x] On-hardware smoke: `insmod datadev.ko` on a target board and confirm a TX path that sets the continue flag actually delivers a continued frame (the bug fix changes observable hardware behavior — worth a quick bench check before merge if a board is available)

### DIFF
--- a/common/driver/axis_gen2.c
+++ b/common/driver/axis_gen2.c
@@ -597,7 +597,6 @@ err_free_rdqueue:
    if (hwData->desc128En) dmaQueueFree(&hwData->rdQueue);
 err_free_wrqueue:
    if (hwData->desc128En) dmaQueueFree(&hwData->wrQueue);
-err_free_hwdata:
    kfree(hwData);
    dev->hwData = NULL;
    return -ENOMEM;
--- a/common/driver/axis_gen2.c
+++ b/common/driver/axis_gen2.c
@@ -176,7 +176,7 @@ inline void AxisG2_WriteTx(struct DmaBuffer *buff, __iomem struct AxisG2Reg *reg
 
    // Configure buffer flags for transmission
    rdData[0]  = (buff->flags >> 13) & 0x00000008;  // bit[3] = continue = flags[16]
-   rdData[0]  = (buff->flags >> 13) & 0x00000010;  // bit[4] = timeout = flags[17]
+   rdData[0] |= (buff->flags >> 13) & 0x00000010;  // bit[4] = timeout = flags[17]
    rdData[0] |= (buff->flags <<  8) & 0x00FF0000;  // Bits[23:16] = lastUser = flags[15:8]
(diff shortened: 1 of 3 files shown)


ruck314:
Confirmed the `// bit[3] = continue = flags[16]` is now working in HW.  Proceeding to tag release

PR #284, opened 2026-06-15
common/driver/gpu_async: Clear read and write address registers on memory remove

JJL772:
### Description

Fixes an issue where FPGAs (one active, and one or more inactive) end up with the same physical GPU address. It seems like even with dropEn=1 on prbs frames, the firmware still initiates an RDMA transaction, which leads to a GPU software hang.

By clearing these registers, we hope to avoid multi-driven buffers and strange hangs.

The easiest way to reproduce this is to have two FPGAs on a system running InterCardTest (or similar) firmware, enable TxEn=1 on the PrbsTx, and then run rdmaTest on one FPGA, and then the other. When you run rdmaTest against the second FPGA, it will hang or transmit only a few frames with corrupt data.

### DIFF
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -63,18 +63,6 @@ jobs:
         run: |
           find . -name '*.h' -o -name '*.cpp' -o -name '*.c' | xargs cpplint
 
-      - name: Assert gpu_async.c unchanged from main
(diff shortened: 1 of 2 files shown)


ruck314 (approved):

JJL772:
@ruck314 is this supposed to be fetching pre-release instead of main?
image

PR #285, opened 2026-06-18
rdmaTest: Move clearing of the handshake space further down

rclaus2:
Clear the handshake space after the possible data copying to avoid printing an incorrect value for the DMA size word.

### Description
To avoid printing an incorrect value for the DMA size word when `s_dumpBytes` is true (`-x` option), the clearing of the handshake space should be performed later in the code.

### DIFF
--- a/data_dev/app/src/rdmaTest.cu
+++ b/data_dev/app/src/rdmaTest.cu
@@ -388,14 +388,7 @@ static void runSimpleLoop(TestSession& s) {
             }
         }
 
-        /* Clear handshake space on the GPU side ("GPU's doorbell"). */
-        assertOk(cuStreamWriteValue32(s.stream,
-            (CUdeviceptr)s.rxBuffers[curBuff] + 4, 0, 0));
-
-        /* Return the buffer index back to the FPGA side ("FPGA's free list"). */
-        assertOk(cuStreamWriteValue32(s.stream,
-            s.regs.dptr + s.coreRegs->freeListOffset(curBuff), 1, 0));
-
+        // Dump header data when requested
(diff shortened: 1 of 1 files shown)


data_dev/app/src/rdmaTest.cu
```diff
@@ -445,6 +438,14 @@ static void runSimpleLoop(TestSession& s) {
...
         }
 
+        /* Clear handshake space on the GPU side ("GPU's doorbell"). */
+        assertOk(cuStreamWriteValue32(s.stream,
+            (CUdeviceptr)s.rxBuffers[curBuff] + 4, 0, 0));
+
+        /* Return the buffer index back to the FPGA side ("FPGA's free list"). */
+        assertOk(cuStreamWriteValue32(s.stream,
+            s.regs.dptr + s.coreRegs->freeListOffset(curBuff), 1, 0));
+
```
JJL772, line 448: I think we already downloaded the header data on line 341

rclaus2: Perhaps I got out of sync with the various updates, but it looks to me like the memcpy on line 413 is grabbing the cleared data which is being printed on line 416.  My recollection was that was showing up incorrectly.

JJL772: Oh! I didn't expand the whole file in GitHub haha. I thought you were fixing an issue with the print in the `if (s_verbose > 1)` block on line 392, which is using previously saved `hdr` data.

I see the buggy code now. Thanks again for the fix

JJL772 (approved):

PR #292, opened 2026-06-30
common/driver: Check for DataGpuEn (0x428) before initting GpuAsyncCore

JJL772:
### Description

Per Matt's suggestion, we'll check for DataGpuEn before attempting to read the GpuAsyncCore version register. Previously the assumption was that reading registers from GpuAsyncCore when DataGpuEn=0 would always return 0.


ruck314 (approved):
yeah, that's probably the safer thing to do is poll the DataGpuEn before trying to poll the AysyncGPU AXI-Lite end point. 

https://github.com/slaclab/axi-pcie-core/blob/main/python/axipcie/_PcieAxiVersion.py#L243

PR #294, opened 2026-06-30
CI Change to auto create de/rpm packages of the datadev driver

BenjiMencer:
Build a ubuntu22 and 24 compitable .deb and rocky9 compatible .rpm files for each new release of the datadev driver

### Description
This change appends to the github ci pipeline and takes the cpu and gpu versions of the datadev driver and compiles it on an ubuntu22,24 and rocky9 container. This change will create 6 extra artifacts (.deb and .rpm files), which will be used to autodeploy the datadev driver to the correct servers via ansible.

### JIRA
https://jira.slac.stanford.edu/browse/TIDIDIT-69

### DIFF
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -3,10 +3,11 @@
 # ----------------------------------------------------------------------------
 # Complete CI/CD pipeline with sequential phases:
 #  Phase 1: Documentation & Linting
-#  Phase 2: CPU Testing (4 distros in parallel) → waits for Phase 1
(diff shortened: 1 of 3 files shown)


JJL772 (approved):
Besides the copilot comments, I don't think I have anything left to add. Thank you for tackling this task!

BenjiMencer:
@JJL772 and @ruck314 I was able to test parts of this code in a forked repo but due to the github secrets I was not able to do a full test, could we run a a quick local test release or a build with a test dummy tag that we delete before merging this into the main release?

PR #300, opened 2026-07-13
common/driver: Indicate unsupported GpuAsyncCore in procfs

JJL772:
### Description

If using an unsupported GpuAsyncCore version, the procfs interface will indicate "GpuAsyncEn 0" which makes it seem like the fw has no GpuAsyncCore support at all.

This PR changes the procfs interface to indicate when a GpuAsyncCore version is unsupported, so you dont need to parse `dmesg` for warnings


ruck314 (approved):
@JJL772 LGTM. merge when ready

PR #301, opened 2026-07-16
datadev: Apply patch to NVIDIA drivers to fix panic-on-exit

JJL772:
### Description

This fixes the crash Mudit was seeing (for real this time...). The previous PR apparently only improved the behavior I guess.

It seems like if you free all p2p pages (as you're supposed to...) the `devmem` struct can get freed by nvidia-uvm and set to NULL. Normally I wouldn't apply a simple NULL check like this, since it is rarely the fix, however the 610 series also adds a NULL check here. So that was indeed the bug.

Unfortunately we only have up to 595 from NVIDIA's own RPM repo right now, so this patch has been included in the meantime. Alternatively, we could check out a 610 version to /usr/local/src and always build with NVIDIA_PATH=/usr/local/src/... while we wait.

The especially strange thing is it seems like this crash is pretty sensitive to *how* the application is killed. For example, if I spin up 6 instances of rdmaTest in the same shell as jobs, and kill them with `kill PID`, no crash. But if I have two login shells each with one instance of rdmaTest, killing either one with ctrl+c (or letting it exit normally) will kill the system.

### DIFF
--- a/data_dev/driver/comp_and_load_drivers.sh
+++ b/data_dev/driver/comp_and_load_drivers.sh
@@ -8,7 +8,8 @@ if [ "$EUID" -ne 0 ]; then
 fi
 
 # Change directory to where this script is
-cd "$(dirname "${BASH_SOURCE[0]}")" || exit 1
+SCRIPTDIR="$(readlink -f "$(dirname "${BASH_SOURCE[0]}")")"
+cd "${SCRIPTDIR}" || exit 1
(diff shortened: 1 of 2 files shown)


ruck314 (approved):
@JJL772 Another than considering adding the hooks for a potential future version that you could case on to not patch this in the future, look fine. Merge when you are ready

PR #303, opened 2026-07-20
Exclusive Access of GpuAsyncCore

JJL772:
### Description

Prior to this change, any userspace application could interact with the GpuAsyncCore state, leading to (what should be) UB in other applications.

At one point I accidentally launched two instances of rdmaTest on the same datadev device, which led to one calling `gpuRemNvidiaMemory` which then clobbered the GpuAsyncCore state under the nose of the other app.


ruck314 (approved):
@JJL772 LGTM.

PR #304, opened 2026-07-27
common/driver: Expose UltraScale SysMon sensors to hwmon interface

JJL772:
### Description

This makes some sensors in the UltraScale SysMon IP block appear under /sys/class/hwmon, and in commands such as 'sensors' (from the lm-sensors package).

### Details

Memory map versioning (here: https://github.com/slaclab/axi-pcie-core/pull/212) is used to determine if AxiSysMon is present or not. This versioning is compatible with older firmware versions, but assumes the sysmon IP block is not present because it was only added in ~2022 or so.

### DIFF
--- a/common/driver/axi_hwmon.c
+++ b/common/driver/axi_hwmon.c
@@ -405,7 +405,6 @@ static int AxiHwmon_Read(struct device *dev, enum hwmon_sensor_types type,
    }
 
    raw = readl(device->base + pvt->axiSysmonOffset + offset);
-   dev_info(device->device, "readl(0x%lX + 0x%lX) = %u\n", pvt->axiSysmonOffset, offset, raw);
 
    // Do conversions if necessary
    if (sens->convFunc)
--- a/common/driver/axi_hwmon.c
+++ b/common/driver/axi_hwmon.c
@@ -18,6 +18,7 @@
 #include <linux/types.h>
 #include <linux/hwmon.h>
 #include <linux/math.h>
+#include <linux/slab.h>
 #include <axi_hwmon.h>
 #include <dma_common.h>
 
--- a/common/driver/axi_hwmon.c
+++ b/common/driver/axi_hwmon.c
@@ -15,14 +15,14 @@
  * contained in the LICENSE.txt file.
  * ----------------------------------------------------------------------------
 **/
-#include "axi_hwmon.h"
-#include "dma_common.h"
 #include <linux/types.h>
 #include <linux/hwmon.h>
 #include <linux/math.h>
+#include <axi_hwmon.h>
+#include <dma_common.h>
 
 /**
- * Specific hardware types for the FPGA. This is used to choose specific ADC 
+ * Specific hardware types for the FPGA. This is used to choose specific ADC
  * conversion functions.
  */
 typedef enum {
@@ -46,17 +46,16 @@ static long convSYSMONE4_T(long raw);
 static long convSYSMONEx_T(struct AxiHwmonDrvPvt* pvt, long raw);
 static long convPS_ADC(struct AxiHwmonDrvPvt* pvt, long raw);
 
-static int AxiHwmon_Read(struct device *dev, enum hwmon_sensor_types type, 
+static int AxiHwmon_Read(struct device *dev, enum hwmon_sensor_types type,
                          u32 attr, int channel, long *val);
 static int AxiHwmon_ReadString(struct device *dev, enum hwmon_sensor_types type,
-		                         u32 attr, int channel, const char **str);
+                               u32 attr, int channel, const char **str);
 static umode_t AxiHwmon_IsVisible(const void *drvdata, enum hwmon_sensor_types type,
-			                         u32 attr, int channel);
+                                  u32 attr, int channel);
 
-enum AxiHwmonSensorInstance {   
-
+enum AxiHwmonSensorInstance {
    AXI_HWMON_CHIP = 0, /* Meta for hwmon, not real sensor */
-   
+
    /**
     * Temperature sensors.
     */
@@ -76,7 +75,7 @@ typedef enum {
    AXI_HWMON_TYPE_C, /* Chip */
    AXI_HWMON_TYPE_T, /* Temp */
    AXI_HWMON_TYPE_V, /* Voltage */
-   
+
    AXI_HWMON_TYPE_COUNT,
 } AxiHwmonSensorType_t;
 
@@ -97,7 +96,7 @@ typedef struct {
 /**
  * Sensor classes.
  */
-const static struct hwmon_channel_info* hwmon_ultrascale_channels[] = {
+static const struct hwmon_channel_info* hwmon_ultrascale_channels[] = {
    HWMON_CHANNEL_INFO(chip, HWMON_C_REGISTER_TZ | HWMON_C_UPDATE_INTERVAL),
    HWMON_CHANNEL_INFO(temp, AXI_HWMON_TEMP_FLAGS),
    HWMON_CHANNEL_INFO(in, AXI_HWMON_IN_FLAGS),
@@ -168,38 +167,22 @@ static const AxiHwmonSensor_t hwmon_ultrascale_sensors[][hwmon_max] = {
 };
 
 /**
- * @brief Maps a AxiHwmonSensorType_t to the equivalent hwmon subsystem type enum
- */
-static enum hwmon_sensor_types sensorTypeToHwmon(AxiHwmonSensorType_t type) {
-   switch(type) {
-   case AXI_HWMON_TYPE_C:
-      return hwmon_chip;
-   case AXI_HWMON_TYPE_T:
-      return hwmon_temp;
-   case AXI_HWMON_TYPE_V:
-      return hwmon_in;
-   default:
-      return hwmon_max;
-   }
-}
-
-/**
  * Initialize Hwmon sensors for this datadev device.
  * If the device is not supported, this prints a warning and returns. This function returns void
  * because it isn't considered critical in datadev initialization. We'll just eat the error and
  * continue anyways.
  * @param dev The datadev device to init for
  * @param axiVerOffset Offset of the AxiVersion register space
- * @param axiSysMonOffset Offset of the AxiSysMonUltrascale register space 
+ * @param axiSysMonOffset Offset of the AxiSysMonUltrascale register space
  */
 void AxiHwmon_Init(struct DmaDevice* dev, off_t axiVerOffset, off_t axiSysMonOffset) {
    AxiHwmonHwType_t hwtype;
 
    // Check for support
    switch (readl(dev->base + axiVerOffset + 0x40C)) {
-   case 0x0: // UltraScale/UltraScale+
+   case 0x0:  // UltraScale/UltraScale+
       break;
-   case 0x1: // 7SERIES (Unsupported)
+   case 0x1:  // 7SERIES (Unsupported)
       return;
    default:
       dev_info(dev->device, "AxiHwmon_Init: No SysMon IP block: hwmon integration will be disabled for this device\n");
@@ -207,7 +190,7 @@ void AxiHwmon_Init(struct DmaDevice* dev, off_t axiVerOffset, off_t axiSysMonOff
    }
 
    // Unfortunately need to case on this. 0x40C does not differentiate between US and US+
-   switch(readl(dev->base + axiVerOffset + 0x424)) {
+   switch (readl(dev->base + axiVerOffset + 0x424)) {
    case 0x00000006:  // XilinxAlveoU50
    case 0x00000007:  // XilinxAlveoU200
    case 0x00000008:  // XilinxAlveoU250
@@ -235,7 +218,7 @@ void AxiHwmon_Init(struct DmaDevice* dev, off_t axiVerOffset, off_t axiSysMonOff
 
    case 0x00000005:  // XilinxAc701
    case 0x0000000A:  // XilinxKc705
-      return; // Unsupported
+      return;  // Unsupported
 
    default:
       dev_warn(dev->device, "AxiHwmon_Init: Unknown PCIe hardware type\n");
@@ -259,8 +242,7 @@ void AxiHwmon_Init(struct DmaDevice* dev, off_t axiVerOffset, off_t axiSysMonOff
       "datadev",
       pvt,
       &hwmon_ultrascale_info,
-      NULL
-   );
+      NULL);
 
    // Check for error and clean up
    if (IS_ERR(pvt->hwmonDev)) {
@@ -363,7 +345,7 @@ static int AxiHwmon_Read(struct device *dev, enum hwmon_sensor_types type,
       dev_warn(dev, "AxiHwmon_Read: Invalid sensor: chan=%d, type=%d, attr=%d\n", channel, type, attr);
       return -ENOTSUPP;
    }
-   
+
    // Lookup the sensor based on the channel and type
    const AxiHwmonSensor_t* sens = &hwmon_ultrascale_sensors[type][channel];
 
@@ -388,7 +370,7 @@ static int AxiHwmon_Read(struct device *dev, enum hwmon_sensor_types type,
       needsHighest = (attr == hwmon_temp_highest);
       needsLowest = (attr == hwmon_temp_lowest);
       break;
-   case hwmon_in: // Voltage, despite the strange name
+   case hwmon_in:  // Voltage, despite the strange name
       needsValue = (attr == hwmon_in_input);
       needsMax = (attr == hwmon_in_max);
       needsCrit = (attr == hwmon_in_crit);
@@ -399,9 +381,9 @@ static int AxiHwmon_Read(struct device *dev, enum hwmon_sensor_types type,
       dev_warn(dev, "AxiHwmon_Read: Unknown sensor type\n");
       return -EOPNOTSUPP;
    }
-   
+
    offset = 0x0;
-   
+
    // Pick the correct offset
    if (needsValue)
       offset = sens->offset;
@@ -444,7 +426,7 @@ static int AxiHwmon_Read(struct device *dev, enum hwmon_sensor_types type,
  * @returns 0 on success, non-zero otherwise.
  */
 static int AxiHwmon_ReadString(struct device *dev, enum hwmon_sensor_types type,
-		                         u32 attr, int channel, const char **str)
+                               u32 attr, int channel, const char **str)
 {
    const char* ptr = NULL;
    const int typeCount = hwmon_ultrascale_chan_counts[type];
@@ -466,7 +448,7 @@ static int AxiHwmon_ReadString(struct device *dev, enum hwmon_sensor_types type,
    default:
       break;
    }
-   
+
    if (ptr)
       *str = ptr;
    else
@@ -479,7 +461,7 @@ static int AxiHwmon_ReadString(struct device *dev, enum hwmon_sensor_types type,
  * Always 0444 in our case because they're read-only.
  */
 static umode_t AxiHwmon_IsVisible(const void *drvdata, enum hwmon_sensor_types type,
-			                         u32 attr, int channel)
+                                  u32 attr, int channel)
 {
    return 0444;
 }
--- a/common/driver/axi_hwmon.c
+++ b/common/driver/axi_hwmon.c
@@ -58,8 +58,6 @@ static umode_t AxiHwmon_IsVisible(const void *drvdata, enum hwmon_sensor_types t
                                   u32 attr, int channel);
 
 enum AxiHwmonSensorInstance {
-   AXI_HWMON_CHIP = 0, /* Meta for hwmon, not real sensor */
-
    /**
     * Temperature sensors.
     */
@@ -93,7 +91,6 @@ typedef struct {
  * Sensor classes.
  */
 static const struct hwmon_channel_info* hwmon_ultrascale_channels[] = {
-   HWMON_CHANNEL_INFO(chip, HWMON_C_REGISTER_TZ),
    HWMON_CHANNEL_INFO(temp, AXI_HWMON_TEMP_FLAGS),
    HWMON_CHANNEL_INFO(in, AXI_HWMON_IN_FLAGS),
    NULL,
--- a/common/driver/axi_hwmon.c
+++ b/common/driver/axi_hwmon.c
@@ -1,0 +1,485 @@
+/**
+ *-----------------------------------------------------------------------------
+ * Company    : SLAC National Accelerator Laboratory
+ *-----------------------------------------------------------------------------
+ * Description: Linux hwmon interface for UltraScale devices implementing the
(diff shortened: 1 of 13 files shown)


common/driver/axi_version.h
```diff
@@ -26,6 +26,7 @@
 
 // Dword offset relative to the start of userValues
 static const int AxiVersion_Usr_HardwareType_Offset = 0x9;
+static const int AxiVersion_Usr_MemoryMapVer_Offset = 0xB;
```
ruck314: @JJL772 Please change from 0xB (11) to 0x3F (63)
https://github.com/slaclab/axi-pcie-core/pull/213

common/driver/data_dev_top.c
```diff
@@ -385,6 +407,11 @@ int DataDev_Probe(struct pci_dev *pcidev, const struct pci_device_id *dev_id) {
       goto err_unmap;
    }
 
+   // Register hwmon interface, if supported.
+   if (axiSysMonOffset != INVALID_REG_OFFSET) {
```
ruck314: @JJL772 The driver should change this `if` state to ...

```
if (axiSysMonOffset != INVALID_REG_OFFSET)  && (GenSysmon != 0) {
```
.. by checking the GenSysmon register: https://github.com/slaclab/axi-pcie-core/blob/pre-release/python/axipcie/_PcieAxiVersion.py#L251

JJL772:
Applied all the feedback and tested against new firmware (and old fw). behaves as expected. Sorry for the delay.

common/driver/axi_hwmon.c
```diff
@@ -0,0 +1,467 @@
...
+ * @param type Sensor type
+ * @param attr The attribute (e.g. hwmon_temp_xxx)
+ * @param channel The channel index for the sensor type
+ * @param val The output value. This will be in mC or mV, depending on the type.
+ * @returns 0 on success, non-zero otherwise
+ */
+static int AxiHwmon_Read(struct device *dev, enum hwmon_sensor_types type,
+                         u32 attr, int channel, long *val)
+{
+   uint32_t value, raw;
```
ruck314: **Blocking.** `value` is `uint32_t` but the conversion functions return `long` and are legitimately negative. `convSYSMONEx_T()` returns `-273820` for `raw == 0` (SysMon not yet converted, or the block held in reset), and any sub-zero ambient gives a negative result too. Truncating at line 411 and widening back at line 415 turns that into `4294693476`, so `sensors` reports about 4.29 million degrees instead of a negative temperature.

```c
long value;
uint32_t raw;
```

common/driver/axi_hwmon.c
```diff
@@ -0,0 +1,467 @@
...
+   case 0x00000006:  // XilinxAlveoU50
+   case 0x00000007:  // XilinxAlveoU200
+   case 0x00000008:  // XilinxAlveoU250
+   case 0x00000009:  // XilinxAlveoU280
+   case 0x0000000C:  // XilinxKcu116
+   case 0x0000000E:  // XilinxVcu128
+   case 0x0000000F:  // XilinxAlveoU55C
+   case 0x00000010:  // XilinxVariumC1100
+   case 0x00001010:  // XilinxVariumC1100Extended
+   case 0x00000012:  // AbacoPc821Ku115
```
ruck314: **Blocking.** `AbacoPc821Ku115` is Kintex UltraScale (SYSMONE1), not UltraScale+. Note that `XilinxKcu1500` at line 212 is also a KU115 part and is correctly placed in the UltraScale group, so these two entries contradict each other.

Using the SYSMONE4 transfer function on a SYSMONE1 part costs about 2.5 C: a raw code corresponding to 50.0 C reads back as roughly 47.4 C.

The full set of Kintex UltraScale (SYSMONE1) hardware types is:

```
0x00_00_00_01  AlphaDataKu3
0x00_00_00_04  SlacPgpCardG4
0x00_00_00_0B  XilinxKcu105
0x00_00_00_0D  XilinxKcu1500
0x00_00_10_0D  XilinxKcu1500Extended
0x00_00_00_11  AbacoPc821Ku085
0x00_00_00_12  AbacoPc821Ku115
```

So `0x00000012` should move to the `AXI_HWMON_HW_TYPE_ULTRASCALE` case.

Note also that only `XilinxKcu1500` and `XilinxVariumC1100` have their bifurcated encoding enumerated. Every other board's (`0x1001`, `0x1004`, `0x100B`, `0x1011`, `0x1012`, and the UltraScale+ equivalents) is unhandled. See the separate comment on line 194.

common/driver/axi_hwmon.c
```diff
@@ -0,0 +1,467 @@
...
+      break;
+   case 0x1:  // 7SERIES (Unsupported)
+      return;
+   default:
+      dev_info(dev->device, "AxiHwmon_Init: No SysMon IP block: hwmon integration will be disabled for this device\n");
+      return;
+   }
+
+   // Unfortunately need to case on this. 0x40C does not differentiate between US and US+
+   switch (readl(dev->base + axiVerOffset + 0x424)) {
```
ruck314: **Blocking.** `userValues[9]` packs the hardware type in bits [11:0] and the bifurcation index in bits [15:12]. `axi_version.c:185-186` already decodes it that way:

```c
htype = aVer->userValues[AxiVersion_Usr_HardwareType_Offset] & 0xFFF;
bifurcationIndex = (aVer->userValues[AxiVersion_Usr_HardwareType_Offset] >> 12) & 0xF;
```

Switching on the unmasked 32-bit word means the two hardcoded `*Extended` encodings (`0x00001010` at line 203, `0x0000100D` at line 212) are just bifurcation index 1 for `XilinxVariumC1100` and `XilinxKcu1500`. Only those two boards get their bifurcated encoding enumerated. Every other card with a nonzero bifurcation index (`0x1001`, `0x1004`, `0x100B`, `0x1011`, `0x1012`, and the UltraScale+ equivalents) falls through to `default:` and logs "Unknown PCIe hardware type" with hwmon silently disabled.

Suggest masking and dropping the two `*Extended` cases:

```c
switch (readl(dev->base + axiVerOffset + 0x400 + 4 * AxiVersion_Usr_HardwareType_Offset) & 0xFFF) {
```

That also removes the magic `0x424` in favor of the existing named constant.

common/driver/axi_hwmon.c
```diff
@@ -0,0 +1,467 @@
...
+         .convFunc = convPS_ADC,
+         .highestOffset = 0x48C,
+         .lowestOffset = 0x49C,
+      },
+   },
+   [hwmon_temp] = {
+      [AXI_HWMON_TEMP] = {
+         .label = "Temperature",
+         .offset = 0x400,
+         .maxOffset = 0x55C,
```
ruck314, line 153: **Blocking.** 0x55C is `OTLowerThresholdRaw` in surf's `_AxiSysMonUltraScale.py`, the over-temp shutdown reset point, not a maximum. `critOffset = 0x54C` (OT upper) is correct.

hwmon's `temp1_max` is meant to be the upper alarm limit, which lives at 0x540 (DRP 0x50, temperature upper alarm). Two options:

- point `maxOffset` at `0x540`, or
- keep 0x55C but expose it as `temp1_crit_hyst` via `HWMON_T_CRIT_HYST`, which is semantically what an OT-lower / reset threshold is.

Also worth noting for `critOffset`: bits [3:0] of the OT upper register are the automatic-shutdown enable pattern rather than temperature data, so the conversion picks up about 0.02 C of error. Negligible, but a `& 0xFFF0` would make the intent explicit.

common/driver/axi_hwmon.c
```diff
@@ -0,0 +1,467 @@
...
+
+#define AXI_HWMON_IN_FLAGS (HWMON_I_INPUT | HWMON_I_LABEL | HWMON_I_HIGHEST | HWMON_I_LOWEST)
+#define AXI_HWMON_TEMP_FLAGS (HWMON_T_INPUT | HWMON_T_MAX | HWMON_T_CRIT | HWMON_T_LABEL \
+                              | HWMON_T_HIGHEST | HWMON_T_LOWEST)
+
+/**
+ * Sensor classes.
+ */
+static const struct hwmon_channel_info* hwmon_ultrascale_channels[] = {
+   HWMON_CHANNEL_INFO(chip, HWMON_C_REGISTER_TZ | HWMON_C_UPDATE_INTERVAL),
```
ruck314: **Blocking.** `HWMON_C_UPDATE_INTERVAL` creates an `update_interval` sysfs file, and `AxiHwmon_IsVisible` returns 0444 unconditionally so it is world readable. But `AxiHwmon_Read` has no `hwmon_chip` case: `hwmon_ultrascale_chan_counts[hwmon_chip]` is 0, so every read takes the guard at line 344, returns `-ENOTSUPP`, and emits an unratelimited `dev_warn`. Any monitoring agent polling sysfs, or a plain `grep . /sys/class/hwmon/hwmon*/*`, spams the kernel log from an unprivileged read.

Suggest dropping `HWMON_C_UPDATE_INTERVAL` (there is no update interval to report, readings are live register reads), and using `dev_warn_ratelimited` on the user-triggerable paths in `AxiHwmon_Read` and `AxiHwmon_ReadString`.

`HWMON_C_REGISTER_TZ` is also not buying anything here: it makes the hwmon core call `hwmon_thermal_add_sensor()` per temp channel, which needs an OF thermal zone. A PCIe card has no DT node, so it fails with `-ENODEV` and just logs "temp1_input not attached to any thermal zone" at probe.

ruck314 (commented):
Reviewed the full diff. Direction looks right and the register-map abstraction is a good idea. CI is green across all ten build and test cells, both earlier review comments are addressed, packaging (symlinks, Makefile objs, DKMS) follows existing conventions, and the teardown ordering is correct: `AxiHwmon_Remove()` runs before `Dma_Clean()` unmaps the register space, and there is no error path after `AxiHwmon_Init()` in probe, so nothing leaks.

Verified the conversion constants against UG580: SYSMONE1 at 502.9098 / 273.8195 and SYSMONE4 at 507.5921 / 279.42636 are both correct for the internal reference, and the 3.0 V supply scaling is right.

Five items I would like fixed before merge, inline below:

1. `AxiHwmon_Read`: `uint32_t value` truncates the signed conversion result, so a zero or sub-zero reading reports as roughly 4.29 million degrees.
2. `AbacoPc821Ku115` (0x12) is classified as UltraScale+, but it is a SYSMONE1 part. This contradicts the `XilinxKcu1500` entry, which is the same silicon.
3. The hardware type switch reads `userValues[9]` unmasked, so cards with a nonzero bifurcation index other than the two hardcoded `*Extended` encodings fall through to "Unknown PCIe hardware type".
4. `temp1_max` points at 0x55C (OT lower threshold), not the temperature upper alarm limit at 0x540.
5. `update_interval` is exported read-only but has no `hwmon_chip` handler, so every read fails and emits an unratelimited `dev_warn`.

The rest are non-blocking.

One item that could not be anchored inline because the file is not in the diff: `docs/explanation/architecture.rst:97-104` quotes the `AGEN2_OFF / PHY_OFF / AVER_OFF / PROM_OFF / USER_OFF` block from `data_dev_top.h` that this PR deletes. It should be updated to describe the versioned map in `axi_pcie_regmap.c`, including which version places AxiSysMon at 0x24000.

common/driver/axi_hwmon.c
```diff
@@ -0,0 +1,467 @@
...
+ * of this distribution and at:
+ *    https://confluence.slac.stanford.edu/display/ppareg/LICENSE.html.
+ * No part of the aes_stream_drivers package, including this file, may be
+ * copied, modified, propagated, or distributed except according to the terms
+ * contained in the LICENSE.txt file.
+ * ----------------------------------------------------------------------------
+**/
+#include <linux/types.h>
+#include <linux/hwmon.h>
+#include <linux/math.h>
```
ruck314: `linux/math.h` is not used anywhere in this file (no `DIV_ROUND_*`, `abs()`, `clamp()`, `mult_frac()`), and it only exists from 5.13 onward. `docs/explanation/kernel-compatibility.rst:4` states the codebase targets 3.10 and up, so this include alone breaks any pre-5.13 build. CI only exercises 5.14+ (Ubuntu 22.04/24.04, Rocky 9, Debian experimental, Fedora rawhide), so nothing catches it. Please drop the include.

Separate but related: `HWMON_CHANNEL_INFO` needs 5.1+ and `hwmon_device_register_with_info` needs 4.9+, so this file cannot compile on the 3.10-4.x end of the documented range regardless. Either guard the hwmon path on `LINUX_VERSION_CODE` the way `dma_common.c` and `dma_buffer.c` do, or update the compatibility doc to state the real floor for `datadev`.

common/driver/data_dev_top.c
```diff
@@ -174,6 +177,7 @@ int DataDev_Probe(struct pci_dev *pcidev, const struct pci_device_id *dev_id) {
    int32_t x;
    uint32_t axiWidth;
    int ret;
+   uint32_t axiGen2Offset, phyOffset, axiSysMonOffset, gpuAsyncCoreOffset;
```
ruck314: `gpuAsyncCoreOffset` is only consumed inside `#ifdef DATA_GPU`, so the default CPU build warns. Reproduced against 6.8.0 headers with `make W=1`:

```
data_dev_top.c:180:56: warning: variable 'gpuAsyncCoreOffset' set but not used [-Wunused-but-set-variable]
```

Suggest moving the declaration and the `AxiRegMap_GetOffset(dev, REG_GPU_ASYNC)` assignment inside the existing `#ifdef DATA_GPU` block.

common/driver/axi_hwmon.c
```diff
@@ -0,0 +1,467 @@
...
+    * Voltage sensors.
+    */
+   AXI_HWMON_VCCINT = 0,
+   AXI_HWMON_VCCAUX,
+   AXI_HWMON_VCCBRAM,
+   AXI_HWMON_IN_COUNT,
+};
+
+typedef enum {
+   AXI_HWMON_TYPE_C, /* Chip */
```
ruck314: `AxiHwmonSensorType_t` and `AXI_HWMON_TYPE_C/T/V/COUNT` are never referenced anywhere in the file. Looks like a leftover from an earlier iteration; suggest deleting.

common/driver/axi_hwmon.c
```diff
@@ -0,0 +1,467 @@
...
+                         u32 attr, int channel, long *val)
+{
+   uint32_t value, raw;
+   uintptr_t offset;
+   int needsValue = 0, needsMax = 0, needsCrit = 0, needsHighest = 0, needsLowest = 0;
+
+   const int typeCount = hwmon_ultrascale_chan_counts[type];
+   if (typeCount == 0 || channel >= typeCount || channel < 0) {
+      dev_warn(dev, "AxiHwmon_Read: Invalid sensor: chan=%d, type=%d, attr=%d\n", channel, type, attr);
+      return -ENOTSUPP;
```
ruck314: `ENOTSUPP` is a kernel-internal errno that is not supposed to reach userspace; checkpatch flags it with "ENOTSUPP is not a SUSV4 error code, prefer EOPNOTSUPP". This file already uses `-EOPNOTSUPP` at lines 383 and 455, so this site and line 435 are inconsistent as well. Suggest `-EOPNOTSUPP` throughout.

common/driver/axi_hwmon.c
```diff
@@ -0,0 +1,467 @@
...
+   }
+
+   // Lookup the sensor based on the channel and type
+   const AxiHwmonSensor_t* sens = &hwmon_ultrascale_sensors[type][channel];
+
+   struct AxiHwmonDrvPvt* pvt = dev->driver_data;
```
ruck314, line 347: Prefer the accessor: `struct AxiHwmonDrvPvt* pvt = dev_get_drvdata(dev);`. Direct `dev->driver_data` access works today but bypasses the documented interface.

common/driver/axi_hwmon.c
```diff
@@ -0,0 +1,467 @@
...
+static const int hwmon_ultrascale_chan_counts[hwmon_max] = {
+   [hwmon_in] = AXI_HWMON_IN_COUNT,
+   [hwmon_temp] = AXI_HWMON_TEMP_COUNT,
+};
+
+/**
+ * Big Ol' array of sensors supported by this device.
+ * Nested based on sensor type, because the hwmon API is not amazing...
+ */
+static const AxiHwmonSensor_t hwmon_ultrascale_sensors[][hwmon_max] = {
```
ruck314, line 125: The inner dimension here is the channel index, but it is sized with `hwmon_max`, which is the number of sensor *types* in `enum hwmon_sensor_types`. It works only because `hwmon_max` happens to be larger than the 3 channels `hwmon_in` needs. A named constant (for example `AXI_HWMON_MAX_CHANNELS`) would say what is meant and would not silently overflow if a future sensor class grows past `hwmon_max` channels.

The bounds checks in `AxiHwmon_Read` and `AxiHwmon_ReadString` do guard the outer index correctly via `hwmon_ultrascale_chan_counts[type] == 0`, so there is no live out-of-bounds read today.

common/driver/axi_hwmon.c
```diff
@@ -0,0 +1,467 @@
...
+   case hwmon_temp:
+      needsValue = (attr == hwmon_temp_input);
+      needsMax = (attr == hwmon_temp_max);
+      needsCrit = (attr == hwmon_temp_crit);
+      needsHighest = (attr == hwmon_temp_highest);
+      needsLowest = (attr == hwmon_temp_lowest);
+      break;
+   case hwmon_in:  // Voltage, despite the strange name
+      needsValue = (attr == hwmon_in_input);
+      needsMax = (attr == hwmon_in_max);
```
ruck314: `hwmon_in_max` and `hwmon_in_crit` are not in `AXI_HWMON_IN_FLAGS`, and none of the voltage sensors set `maxOffset` / `critOffset`, so these two branches are unreachable. Harmless, but they suggest coverage that does not exist.

common/driver/axi_hwmon.c
```diff
@@ -0,0 +1,467 @@
...
+   // Check for error and clean up
+   if (IS_ERR(pvt->hwmonDev)) {
+      dev_err(dev->device, "AxiHwmon_Init: Failed to register hwmon device: %ld\n",
+              PTR_ERR(pvt->hwmonDev));
+      goto error_post_alloc;
+   }
+
+   dev_info(dev->device, "AxiHwmon_Init: Registered hwmon device\n");
+
+   dev->hwmonPrivate = pvt;
```
ruck314: `dev->hwmonPrivate` is assigned after `hwmon_device_register_with_info()` returns, so the sysfs attributes are live for a short window before the field is set. Nothing reads `hwmonPrivate` on that path today (the read callbacks go through drvdata), so this is not a live bug, but setting it before registering would remove the window.

common/driver/axi_pcie_regmap.c
```diff
@@ -0,0 +1,124 @@
...
+ */
+int AxiRegMap_Init(struct DmaDevice* dev, __iomem void* axiVersion) {
+   __iomem struct AxiVersion_Reg *reg = (__iomem struct AxiVersion_Reg *)axiVersion;
+   dev->regMapVersion = readl(&reg->userValues[AxiVersion_Usr_MemoryMapVer_Offset]) & 0xFF;
+
+   if (dev->regMapVersion > MAX_SUPPORTED_REG_VERSION) {
+      dev_warn(dev->device, "AxiRegMap_Init: Unsupported register map version %u. Maximum supported version is %d.\n",
+               dev->regMapVersion, MAX_SUPPORTED_REG_VERSION);
+      dev_warn(dev->device, "Please use a newer driver version, or contact a maintainer to update the software\n");
+      return -1;
```
ruck314: Two things here:

1. `return -1` propagates to `DataDev_Probe`, which converts it to `-EINVAL`. Returning a proper negative errno directly would be more idiomatic and would let the caller just forward it.
2. Policy check: an unrecognized map version fails the probe outright, so a card running firmware newer than the driver does not come up at all. That is defensible, since guessing at offsets is worse than not loading, and old firmware reads 0 here so backward compatibility holds. Just confirming it is deliberate rather than falling back to V0 with a loud warning.

common/driver/dma_common.h
```diff
@@ -89,6 +89,7 @@ typedef uint32_t __poll_t;
  * @txBuffers: List of transmit buffers.
  * @rxBuffers: List of receive buffers.
  * @tq: Transmit queue structure.
+ * @regMapVersion: The register map version.
```
ruck314, line 92: `hwmonPrivate` was added to the struct at line 167 but is not in this kernel-doc block, while `regMapVersion` is. Suggest adding:

```
 * @hwmonPrivate: Private data for the hwmon interface (axi_hwmon.c).
```

common/driver/axi_hwmon.c
JJL772: This was intentional so the array could be indexed based on the hwmon sensor type, and so knowledge of the underlying enum values of `hwmon_in`/`hwmon_temp` is not needed when defining `hwmon_ultrascale_chan_counts`.

common/driver/axi_pcie_regmap.c
JJL772: Yes this is deliberate and is based on the assumption that the memory map should rarely change.
It seems like most of the historical memory map changes have been fairly minor though and did not require driver changes. So this is introducing slightly more work for those types of changes.

common/driver/axi_hwmon.c
JJL772: Did the rhel7 CI get removed? I expected to see a CI failure if it was not supported

ruck314: Yes, RHEL7 coverage is gone. The old `aes_ci.yml` compile matrix included `ghcr.io/jjl772/centos7-vault` (kernel 3.10), but that workflow was replaced by `ci_pipeline.yml` in cd0d771, and the new matrix is ubuntu:22.04, ubuntu:24.04, rockylinux:9, debian:experimental, fedora:rawhide. The oldest kernel built now is Rocky 9's 5.14.

`include/linux/math.h` first appeared in v5.11 (aa6159ab99a9, "kernel.h: split out mathematical helpers"), so nothing left in the matrix can catch this. It would also break RHEL8 (4.18) and Debian 11 (5.10), and `docs/explanation/kernel-compatibility.rst` still advertises 3.10 as the floor.

Good catch either way, since the include is not actually needed: `axi_hwmon.c` only does plain integer `*` and `/` in the conversion functions, no `DIV_ROUND_*`/`mult_frac`/etc. Dropping it.

common/driver/axi_hwmon.c
JJL772: ~~I'm not sure why 0x55C would be a hysteresis and not something like `hwmon_temp_emergency`.~~ Hysteresis does make sense for this value actually, but I dont think it's worth exposing. I am just going to correct the offset and mask off the lower bits

JJL772: Actually `0x540` does not need masking either

common/driver/axi_hwmon.c
JJL772: Also removed the chip sensor def since it does not appear to be required.

ruck314 (approved):

PR #309, opened 2026-08-07
datadev/app/rdmaTest: Use CUDA driver API and align buffer sizes to 64k (GPU page size)

JJL772:
### Description

The driver enforces a 64k size alignment to ensure it can pin whole pages, so we need to make sure that our buffers are size-aligned accordingly. There are some other risks here I think, like if we allocate `64k + 16` bytes, map that (2 pages mapped = 128k total, even if we only meant 64k+16 bytes), and a subsequent unrelated call to cudaMalloc allocates out of the same page.

the `cuMemAlloc` change was somewhat arbitrary, but required for dma-buf support (in a future PR). `cuMemGetHandleForAddressRange` requires buffers allocated by `cuMemAlloc` or the low level VMM API.

### DIFF
--- a/data_dev/app/src/rdmaTest.cu
+++ b/data_dev/app/src/rdmaTest.cu
@@ -53,7 +53,8 @@ static int str2int(const char* s) {
 /* CUDA Allocations must be aligned to this size */
 #define GPU_PAGE_SIZE 0x10000
 
-#define ALIGN_VALUE(_x, _align) (((_x) + (_align) + 1) & ~((_align)- 1))
+/* Aligns _x to a power-of-two boundary */
+#define ALIGN_VALUE(_x, _align) (((_x) + (_align) - 1) & ~((_align) - 1))
 
 /**
  * @brief Per-session state. Replaces the old GpuAsyncContext lifecycle wrapper
@@ -238,7 +239,7 @@ static int initSession(TestSession& s, const char* dev, int gpuIdx,
         s.txBuffers.resize(bufCnt, 0);
         for (int i = 0; i < bufCnt; ++i) {
             if (cuMemAlloc(&s.txBuffers[i], ALIGN_VALUE(bufSize + s.dmaHeaderSize, GPU_PAGE_SIZE)) != CUDA_SUCCESS) {
-                fprintf(stderr, "cuMalloc(rxBuffers[%d]) failed\n", i);
+                fprintf(stderr, "cuMalloc(txBuffers[%d]) failed\n", i);
                 return -1;
             }
             if (gpuAddNvidiaMemory(fd, 0, (uint64_t)s.txBuffers[i], bufSize) < 0) {
--- a/data_dev/app/src/rdmaTest.cu
+++ b/data_dev/app/src/rdmaTest.cu
@@ -40,7 +40,6 @@ static int s_dumpBytes = 0;
 static int s_cnt = -1;
 static std::string s_dumpFile;
 
-static void assertOk(cudaError_t err);
 static void assertOk(CUresult err);
 static void showHelp();
 
@@ -50,6 +49,11 @@ static int str2int(const char* s) {
         base = 16;
     return strtol(s, NULL, base);
 }
+
+/* CUDA Allocations must be aligned to this size */
+#define GPU_PAGE_SIZE 0x10000
+
+#define ALIGN_VALUE(_x, _align) (((_x) + (_align) + 1) & ~((_align)- 1))
 
 /**
  * @brief Per-session state. Replaces the old GpuAsyncContext lifecycle wrapper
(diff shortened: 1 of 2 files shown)


data_dev/app/src/rdmaTest.cu
```diff
@@ -51,6 +50,11 @@ static int str2int(const char* s) {
     return strtol(s, NULL, base);
 }
 
+/* CUDA Allocations must be aligned to this size */
+#define GPU_PAGE_SIZE 0x10000
+
+#define ALIGN_VALUE(_x, _align) (((_x) + (_align) + 1) & ~((_align)- 1))
```
ruck314: `+ 1` should be `- 1`. This is the standard round-up-to-a-power-of-two idiom:

```c
#define ALIGN_VALUE(_x, _align) (((_x) + (_align) - 1) & ~((_align) - 1))
```

As written it never under-allocates, so nothing breaks functionally, but it rounds up an extra full page whenever `_x` is already a multiple of `_align` (and also when `_x % _align == _align - 1`):

| requested | this macro | expected |
|---|---|---|
| `0x10000` | `0x20000` | `0x10000` |
| `0x80000` | `0x90000` | `0x80000` |
| `0x100000` | `0x110000` | `0x100000` |

With the defaults (`bufSize = 0x100000`, `bufCnt = 1024`) that is an extra 64 MiB of GPU memory reserved whenever `dmaDataBytes()` reports 0.

Also worth a comment that the macro is only valid for a power-of-two `_align`, since the next person to reach for it will not necessarily check.

data_dev/app/src/rdmaTest.cu
```diff
@@ -231,10 +235,10 @@ static int initSession(TestSession& s, const char* dev, int gpuIdx,
         }
     }
     if (loopback) {
-        s.txBuffers.resize(bufCnt, nullptr);
+        s.txBuffers.resize(bufCnt, 0);
         for (int i = 0; i < bufCnt; ++i) {
-            if (cudaMalloc(&s.txBuffers[i], bufSize + s.dmaHeaderSize) != cudaSuccess) {
-                fprintf(stderr, "cudaMalloc(txBuffers[%d]) failed\n", i);
+            if (cuMemAlloc(&s.txBuffers[i], ALIGN_VALUE(bufSize + s.dmaHeaderSize, GPU_PAGE_SIZE)) != CUDA_SUCCESS) {
+                fprintf(stderr, "cuMalloc(rxBuffers[%d]) failed\n", i);
```
ruck314: This is the `txBuffers` loop, so the message should say `txBuffers`. The pre-change line had it right, so this is a regression introduced here. The two allocation loops are otherwise identical, which makes a failure here point at the wrong one.

While editing the string: `cuMalloc` is not a real API name, `cuMemAlloc` is. Same typo on line 228 and in the step-6 doc comment on line 130.

ruck314:
One design question before this goes in, on the size-alignment rationale.

The failure mode described in the PR body (an unrelated subsequent allocation landing in a page we pinned) is governed by the buffer's *base address*, not its size. From `common/driver/gpu_async.c`:

```c
virt_start  = buffer->address & GPU_BOUND_MASK;
virt_offset = buffer->address & ~GPU_BOUND_MASK;
pin_size    = (buffer->address + buffer->size - virt_start + GPU_BOUND_OFFSET) & GPU_BOUND_MASK;
```

`cuMemAlloc` only promises that the result is "suitably aligned for any kind of variable", not 64k. If the returned pointer is not 64k-aligned, the padded allocation still straddles a partial page at each end, `virt_offset` is nonzero, `pin_size` rounds up to an extra page, and an unrelated allocation can still land in it. So padding the size narrows the window but does not close it.

The usual GPUDirect pattern (gdrcopy, the NVIDIA samples) over-allocates and rounds the pointer instead:

```c
cuMemAlloc(&raw, size + GPU_PAGE_SIZE - 1);
aligned = (raw + GPU_PAGE_SIZE - 1) & ~(CUdeviceptr)(GPU_PAGE_SIZE - 1);
```

`TestSession` would need to keep both, since `cuMemFree` wants the raw pointer.

I am fine either way: implement the pointer alignment, or leave the size padding as-is and reword the commit message so it does not claim a guarantee the code cannot provide. Mostly I want it written down so the dma-buf PR does not inherit the assumption that these buffers are page-aligned.

data_dev/app/src/rdmaTest.cu
JJL772: Whoops!

JJL772:
> cuMemAlloc only promises that the result is "suitably aligned for any kind of variable", not 64k. If the returned pointer is not 64k-aligned, the padded allocation still straddles a partial page at each end, virt_offset is nonzero, pin_size rounds up to an extra page, and an unrelated allocation can still land in it. So padding the size narrows the window but does not close it.

We also enforce alignment of the memory region in the driver. If the address alignment and size alignment are both 64k we should never straddle a GPU page boundary. 

In practice, it seems like `cuMemAlloc`/`cudaMalloc` always give us memory aligned to the nearest power of two of the byte size (thus, 64k blocks are 64k aligned). This is likely relying on some internal driver behavior, so I agree we should change it. If they violate this assumption, the driver will reject with `-EINVAL` so we should not end up with unexpected behavior.

I'm going to refactor this in my separate dma-buf PR since I'm reworking memory allocation there

ruck314 (approved):

JJL772:
Closing in favor of upcoming changes adding dma-buf support (that also changes this code)

PR #313, opened 2026-08-27
Abort datadev probe when the AxiVersion uptime counter reads all-ones

ruck314:
### Description

The `datadev` probe now reads the AxiVersion uptime counter immediately after the register space is mapped and fails with `-EIO` if it reads all-ones. The uptime counter counts seconds since the FPGA left reset, so `0xFFFFFFFF` (2^32 - 1 seconds, roughly 136 years) is not a value real firmware can report. Reading it back means the PCIe read never reached the FPGA and was completed with an abort or a timeout instead, which is what happens when the FPGA is reprogrammed and neither a reboot nor a PCIe rescan was performed afterwards to bring the link back up.

Every register read past that point returns the same all-ones garbage, so without an early check the probe fails much later and for reasons that point away from the actual cause. The new failure message names the real problem and the recovery action:

```
datadev 0000:af:00.0: Init: AxiVersion uptime counter reads all-ones; the PCIe
link to the FPGA is down. Reboot the host or rescan the PCIe bus to recover the
link after reprogramming the FPGA.
```

Also adds `AxiVersion_GetUpTime()` so the counter offset stays expressed through `struct AxiVersion_Reg` rather than a bare constant.

### Details

This came out of a field report on `drp-neh-cmp018`, where a card in this state produced:

```
datadev 0000:af:00.0: Init: Using 255-bit DMA mask.
datadev 0000:af:00.0: Init: Using 255-bit coherent DMA mask.
datadev 0000:af:00.0: Init: dma_alloc_coherent failed for read ring (size=0).
datadev 0000:af:00.0: Init: Card-specific init failed: -12.
```

Both oddities trace back to all-ones register reads:

- `data_dev_top.c` computes `axiWidth = (readl(dev->reg + 0x34) >> 8) & 0xFF`, giving 255. `DMA_BIT_MASK(255)` shifts by more than the operand width, and on x86-64 the count is masked to 6 bits, yielding a 63-bit mask that `dma_set_mask()` accepts. The probe continued instead of stopping.
- `axis_gen2.c` computes `hwData->addrCount = (1 << readl(&(reg->addrWidth)))`. With 255, the 32-bit shift count is masked to 5 bits, giving `0x80000000`; multiplying by 16 for 128-bit descriptors overflows `uint32_t` to exactly 0. `dma_alloc_coherent(dev, 0, ...)` then returns NULL and reports `-ENOMEM`, which reads as a host memory problem rather than a dead link.

On this branch the first misleading symptom would instead come from `AxiRegMap_Init()`, which reads `0xFFFFFFFF & 0xFF` as the register map version and warns "Unsupported register map version 255. Maximum supported version is N / Please use a newer driver version", advising a driver upgrade for what is actually a link problem. The new check is placed between `Dma_MapReg()` and `AxiRegMap_Init()` so it runs before that.
uild_version`, would make it testable.

### DIFF
--- a/common/driver/axi_version.c
+++ b/common/driver/axi_version.c
@@ -214,3 +214,20 @@ void AxiVersion_SetUserReset(__iomem void *base, bool state) {
 
    writel(val, &(reg->userReset));  // Write the value to the userReset register
 }
+
+/**
+ * AxiVersion_GetUpTime - Reads the AXI version uptime counter.
+ * @base: Pointer to the base address of the AXI Version register block.
+ *
+ * Reads the free-running counter that tracks the number of seconds since the
+ * FPGA left reset. Besides reporting uptime, this register doubles as a cheap
+ * liveness check on the register interface: all-ones is not a value real
+ * firmware can report, so it indicates the read never reached the FPGA.
+ *
+ * Return: the raw contents of the uptime counter register.
+ */
+uint32_t AxiVersion_GetUpTime(__iomem void *base) {
+   __iomem struct AxiVersion_Reg *reg = (__iomem struct AxiVersion_Reg *)base;
+
+   return readl(&(reg->upTimeCount));
+}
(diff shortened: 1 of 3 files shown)


rclaus2 (approved):
Looks sensible and good to me.

PR #314, opened 2026-09-03
fix(pgpcard): restore the PGP Gen2/Gen3 driver for Rocky Linux 9

ruck314:
### Description

Restores the `pgpcard` driver so it builds on Rocky Linux 9, for a user with a
legacy PGP-GEN3 card (`1a4a:2020`) on a migrated server. The tree was removed
from `main` in early 2024 and resurrected on this branch, where it did not build.

The breakage was not kernel API churn. `pgpcard/driver/src/dma_common.{c,h}` and
`dma_buffer.{c,h}` are symlinks into `common/driver/`, so the driver already
inherits the modern guard set including RHEL 9 backport detection. What broke was
this repo's own interfaces plus three headers deleted alongside the driver:

- **Restored, pgpcard-local** so future removal touches nothing shared:
  `include/PgpDriver.h` and `FpgaProm.h` (deleted by `37f2178`),
  `app_lib/PciCardProm.{h,cpp}` (deleted by `39fcb5c`) and its `McsRead`
  dependency. `PciCardProm.h` also gains the `<string>` include it had been
  getting by accident from its includer.
- **`hwFunc.init` now returns `int`, not `void`.** A hard error under
  `-Werror=incompatible-pointer-types`. The `PgpInfo` allocation moves above the
  first register write and returns `-ENOMEM`, since `Dma_Init` does not call
  `->clear()` when `->init()` fails. Also fixes an unchecked `kmalloc`.
- **`hwFunc.irqEnable` was NULL in both function tables.** `Dma_Clean` calls it
  guarded only on `hwFunc`, not the member, so removing a probed card
  dereferenced NULL. Adds `PgpCardG2_IrqEnable` and `PgpCardG3_IrqEnable`.

Also adds a `pgp` phase to the local harness and GitHub CI alongside `cpu` and
`gpu`, clears the 707 cpplint errors that were breaking the repo-wide Phase 1
gate for every PR, and consolidates pgpcard docs into `pgpcard/README.md`.

### Details

Builds clean, no warnings, across the datadev distro matrix:

| Cell | Kernel |
|---|---|
| `rockylinux:9` | `5.14.0-687.42.1.el9_8` (genuine RHEL 9, the target) |
| `ubuntu:22.04` | `5.15.0-190-generic` |
| `ubuntu:24.04` | `6.8.0-139-generic` |
| `debian:experimental` | `7.1.12+deb14-amd64` |
| `fedora:rawhide` | `7.3.0-0.rc0…fc46` |

No new version guards were needed. pgpcard-specific code touches only APIs stable
across 3.10 to 7.x, which is why it reached 7.3-rc unmodified, past the 6.8
ceiling in `docs/explanation/kernel-compatibility.rst`.


baljamal:
it tested with Rocky 9.6 which has kernel version 5.14.0-570.17.1.
but I had to make these changes to make it work.

`diff --git a/common/driver/dma_common.c b/common/driver/dma_common.c
index 8aa05cc..df33b22 100755
--- a/common/driver/dma_common.c
+++ b/common/driver/dma_common.c
@@ -148,11 +148,7 @@ struct class *gCl;
  *
  * Returns NULL always.
  */
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 2, 0) || (defined(RHEL_RELEASE_CODE) && RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(9, 4))
 char *Dma_DevNode(const struct device *dev, umode_t *mode) {
-#else
-char *Dma_DevNode(struct device *dev, umode_t *mode) {
-#endif
    if (mode != NULL) {
       *mode = 0666;
    }
diff --git a/common/driver/dma_common.h b/common/driver/dma_common.h
index c4b0e52..c037783 100755
--- a/common/driver/dma_common.h
+++ b/common/driver/dma_common.h
@@ -253,8 +253,6 @@ extern struct file_operations DmaFunctions;
 // Function prototypes
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 2, 0) || (defined(RHEL_RELEASE_CODE) && RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(9, 4))
 char *Dma_DevNode(const struct device *dev, umode_t *mode);
-#else
-char *Dma_DevNode(struct device *dev, umode_t *mode);
 #endif
 int Dma_MapReg(struct DmaDevice *dev);
 int Dma_Init(struct DmaDevice *dev);
`

I think we need to change kernel version in the condition but I don't at which version the function signature changed

ruck314:
@baljamal thanks, you were right that the condition was wrong: the signature changed in RHEL/Rocky **9.3**, not 9.4.

I pulled the `kernel-devel` headers for Rocky 9.0 through 9.8 and read them directly:

- `class.devnode()` const arg: 9.2 (5.14.0-284) is non-const, 9.3 (5.14.0-362) is const. Guard said 9.4.
- `class_create()` dropping the module arg: 9.4. That guard was already correct.
- Same sweep turned up two more off-by-one boundaries: `PDE_DATA` -> `pde_data()` is 9.1 (guard said 9.3), and `vm_flags_set()` is 9.5 (guard said 9.4, it is absent from every 9.4 header).

Fixed in 14b9995. The devnode condition is now a single macro shared by the prototype and the definition, so those two cannot disagree again. `aximemorymap` had already drifted that way (header guarded on 6.4, definition on 6.2), which broke upstream 6.2/6.3 builds.

Verified by building datadev, pgpcard, aximemorymap and the emulator against kernel-devel for Rocky 9.0 through 9.8, plus Fedora 6.2.9 and Ubuntu 6.8. Before the fix: 9.1/9.2 failed on `PDE_DATA`, 9.3 failed on the devnode assignment, 9.4 failed on `vm_flags_set`. All clean now.

One thing I could not explain: against stock 9.6 headers (`RHEL_RELEASE_CODE=2310`) the old code compiles clean, so your patch should have been a no-op there, yet your exact error reproduces on 9.3 headers. Could you post the output of

```
grep RHEL /lib/modules/$(uname -r)/build/include/generated/uapi/linux/version.h
```

so we can see which minor your build actually resolved to? The fix stands either way. Please re-test with your HW when you get a chance.

baljamal:
grep output for ctl-xrt-wave8-01 where IOC lives:

#define RHEL_MAJOR 9
#define RHEL_MINOR 6
#define RHEL_RELEASE_VERSION(a,b) (((a) << 8) + (b))
#define RHEL_RELEASE_CODE 2310
#define RHEL_RELEASE "570.17.1"

grep output for psbuild-rocky9-01 build machine:


#define RHEL_MAJOR 9
#define RHEL_MINOR 6
#define RHEL_RELEASE_VERSION(a,b) (((a) << 8) + (b))
#define RHEL_RELEASE_CODE 2310
#define RHEL_RELEASE "570.58.1"

baljamal:
I built it using your fix and it build clean. I will test with hardware and get back to you

baljamal:
I think the answer for your question is RHEL backporting of an upstream Linux API change

baljamal:
that change made in commit ff62b8e which got merged to kernel 6.2+ then backported to RHEL 9.4+

ruck314:
Right mechanism, off by one minor. `ff62b8e6588f` landed in **9.3**, not 9.4. Straight from the `kernel-devel` headers (`include/linux/device/class.h`):

| Rocky | Kernel | `RHEL_RELEASE_CODE` | `devnode` first arg |
|---|---|---|---|
| 9.2 | 5.14.0-284.30.1 | 2306 | `struct device *` |
| **9.3** | **5.14.0-362.8.1** | **2307** | **`const struct device *`** |
| 9.4 | 5.14.0-427.16.1 | 2308 | `const struct device *` |

So the `>= RHEL_RELEASE_VERSION(9, 3)` guard in 14b9995 matches the headers and stays as is.

That does leave your original failure unexplained, though. Both your hosts report `RHEL_RELEASE_CODE=2310`, which satisfies the old `>= 9.4` guard *and* the new `>= 9.3` one, so `Dma_DevNode` was already `const` for you before my fix. I checked all three corrected boundaries through the preprocessor at 2310 and 14b9995 is a no-op on 9.6. Your patch could not have changed anything there either, so something else broke that build.

The one preprocessor state that matches your symptom is `RHEL_RELEASE_CODE` not being visible at compile time: `LINUX_VERSION_CODE` is 5.14, the guard falls through to non-const, and that collides with the kernel's `const` devnode. My fix does not repair that case either. Could you post the original failing output, or `make -C pgpcard/driver V=1 2>&1 | head -40`? If the macro is genuinely missing in your build, there is still a real bug here to fix.

One thing before you test: `psbuild-rocky9-01` is `570.58.1` and `ctl-xrt-wave8-01` is `570.17.1`. `KERNELDIR` comes from `uname -r`, so a `.ko` built on psbuild carries vermagic `5.14.0-570.58.1.el9_6.x86_64` and the IOC will reject it with `Invalid module format`. Either build on the IOC or pass `KVER=5.14.0-570.17.1.el9_6.x86_64` with the matching `kernel-devel` installed (`pgpcard/README.md`).

baljamal:
before your fix it was building and producing .ko files.

but I had these two messages

```
/cds/group/pcds/epics-dev/baljamal/forks/drivers/aes-stream-drivers/data_dev/driver/src/dma_common.c:277:20: error: assignment to ‘char * (*)(const struct device *, umode_t *)’ {aka ‘char * (*)(const struct device *, short unsigned int *)’} from incompatible pointer type ‘char * (*)(struct device *, umode_t *)’ {aka ‘char * (*)(struct device *, short unsigned int *)’} [-Werror=incompatible-pointer-types]
  277 |       gCl->devnode = Dma_DevNode;
      |                    ^
cc1: some warnings being treated as errors
make[3]: *** [scripts/Makefile.build:299: /cds/group/pcds/epics-dev/baljamal/forks/drivers/aes-stream-drivers/data_dev/driver/src/dma_common.o] Error 1
make[2]: *** [Makefile:1928: /cds/group/pcds/epics-dev/baljamal/forks/drivers/aes-stream-drivers/data_dev/driver] Error 2
make[2]: Leaving directory '/usr/src/kernels/5.14.0-362.18.1.el9_3.0.1.x86_64'
make[1]: *** [Makefile:107: all] Error 2
make[1]: Leaving directory '/cds/group/pcds/epics-dev/baljamal/forks/drivers/aes-stream-drivers/data_dev/driver'
cp: cannot stat '/cds/group/pcds/epics-dev/baljamal/forks/drivers/aes-stream-drivers/data_dev/driver/*.ko': No such file or directory
```

```
/cds/group/pcds/epics-dev/baljamal/forks/drivers/aes-stream-drivers/pgpcard/driver/src/dma_common.c:277:20: error: assignment to ‘char * (*)(const struct device *, umode_t *)’ {aka ‘char * (*)(const struct device *, short unsigned int *)’} from incompatible pointer type ‘char * (*)(struct device *, umode_t *)’ {aka ‘char * (*)(struct device *, short unsigned int *)’} [-Werror=incompatible-pointer-types]
  277 |       gCl->devnode = Dma_DevNode;
      |                    ^
cc1: some warnings being treated as errors
make[3]: *** [scripts/Makefile.build:299: /cds/group/pcds/epics-dev/baljamal/forks/drivers/aes-stream-drivers/pgpcard/driver/src/dma_common.o] Error 1
make[2]: *** [Makefile:1928: /cds/group/pcds/epics-dev/baljamal/forks/drivers/aes-stream-drivers/pgpcard/driver] Error 2
make[2]: Leaving directory '/usr/src/kernels/5.14.0-362.18.1.el9_3.0.1.x86_64'
make[1]: *** [Makefile:68: all] Error 2
make[1]: Leaving directory '/cds/group/pcds/epics-dev/baljamal/forks/drivers/aes-stream-drivers/pgpcard/driver'
cp: cannot stat '/cds/group/pcds/epics-dev/baljamal/forks/drivers/aes-stream-drivers/pgpcard/driver/*.ko': No such file or directory

```

ruck314:
That solves it, and it is the boundary after all. Your log shows the build ran in:

```
make[2]: Leaving directory '/usr/src/kernels/5.14.0-362.18.1.el9_3.0.1.x86_64'
```

That is **el9_3**, `RHEL_RELEASE_CODE=2307`, not the 9.6 your `grep` reported. So:

- old guard `>= RHEL_RELEASE_VERSION(9, 4)` = 2308: 2307 < 2308 picks non-const `Dma_DevNode`, but 9.3's `class.devnode` is already const, giving your exact error at `gCl->devnode = Dma_DevNode;`
- new guard `>= RHEL_RELEASE_VERSION(9, 3)` = 2307: 2307 >= 2307 picks const, compiles

So 14b9995 is precisely the fix for this, and I am retracting my "it is a no-op, something else broke it" call from earlier. That was correct for 9.6 and wrong for you, because I took the `grep` output as the kernel you were building against.

Worth sorting out before the hardware test, though: that host is resolving to 9.3 `kernel-devel` while `ctl-xrt-wave8-01` runs `5.14.0-570.17.1.el9_6`. A `.ko` built there carries vermagic `5.14.0-362.18.1.el9_3.0.1.x86_64` and the IOC will refuse it with `Invalid module format`. Worth confirming with `modinfo -F vermagic pgpcard/driver/pgpcard.ko` before you try loading it, and building either on the IOC or with `KVER=5.14.0-570.17.1.el9_6.x86_64` and the matching `kernel-devel` installed.

PR #317, opened 2026-09-08
gpu_async: Make GpuAsyncCore ownership per process, not per thread

rclaus2:
Make GpuAsyncCore ownership per process, not per thread.

### Description
GpuAsync.h documents the GpuAsyncCore state as "exclusively owned by one process    at a time", but the ownership checks compare against current->pid, which in the kernel is the thread id; current->tgid is what userspace calls the PID.  A multithreaded process that claims ownership with gpuAddNvidiaMemory() on one thread therefore cannot call gpuSetWriteEn(), gpuEnableTx() or gpuEnableRx() from any other thread: the ioctl fails with EBUSY and the driver logs

      datadev 0000:04:00.0: Gpu_SetWriteEn: Called by non-owner PID (3417898)

even though 3417898 is a thread of the owning process.

This is how the LCLS-II GPU DRP is structured: MemPoolGpu's constructor registers the GPU buffers, while the per-device threads that rearm DMA buffers for writing call gpuSetWriteEn(), so every rearm is rejected.

Compare against current->tgid in all five ownership sites, which is what the documented contract calls for and lets any thread of the owning process issue the ioctls.  Also note the semantics at the field declaration, since it stores a TGID rather than the thread id its name suggests.

While here, Gpu_EnableTx() and Gpu_EnableRx() both logged their non-owner warning as "Gpu_SetWriteEn", which misdirects anyone diagnosing this; name the right function in each.

### DIFF
--- a/common/driver/gpu_async.h
+++ b/common/driver/gpu_async.h
@@ -101,7 +101,10 @@ struct GpuData {
    int32_t version;
    uint32_t maxBuffers;
    uint32_t disabled;
-   atomic64_t pid;
+   atomic64_t pid;              // TGID of the owning process, or 0 when unowned.
+                                // Compared against current->tgid, not current->pid:
+                                // ownership is per process, so that any thread may
+                                // issue the GpuAsyncCore ioctls.
    struct GpuBuffers writeBuffers;
    struct GpuBuffers readBuffers;
 };
--- a/common/driver/gpu_async.c
+++ b/common/driver/gpu_async.c
@@ -201,10 +201,10 @@ int32_t Gpu_AddNvidia(struct DmaDevice *dev, uint64_t arg) {
    }
 
    // Check if another PID already owns this GpuAsyncCore state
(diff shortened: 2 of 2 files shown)


rclaus2:
The changes were tested on gpu008 against 5 datadevs.  The build completed cleanly and the DAQ behaved correctly and as expected with the change.

common/driver/gpu_async.h
```diff
@@ -101,7 +101,10 @@ struct GpuData {
    int32_t version;
    uint32_t maxBuffers;
    uint32_t disabled;
-   atomic64_t pid;
+   atomic64_t pid;              // TGID of the owning process, or 0 when unowned.
```
JJL772, line 104: This should be renamed `tgid` (also makes the comment redundant)

JJL772 (approved):
LGTM; thank you for the fix! 

Regarding my review comment, I will apply that change in a future PR.

PR #318, opened 2026-09-09
common/driver: Rename pid -> tgid in GpuData struct

JJL772:
### Description

Minor cleanup, renames pid -> tgid in the GpuData struct for clarity (see my comment on #317)


rclaus2 (approved):
I think this looks fine.

PR #319, opened 2026-09-12
fix(data_dev): make Makefile.local reach the kbuild pass

rclaus2:
data_dev: Fix datadev-gpu DKMS builds silently lacking GPU support

### Description
`datadev-gpu-dkms` has never produced a GPU-enabled module. `-include Makefile.local` in `data_dev/driver/Makefile` used a bare relative path, so it resolved in the top-level pass but not in the kbuild sub-make, whose working directory is `$(KERNELDIR)`. `ccflags-y` is evaluated only in that second pass, so `NVIDIA_DRIVERS` was empty exactly where it mattered and `-DDATA_GPU=1` was never added.

The failure is quiet in a bad way. Because the include *does* resolve in the first pass, `$(info)` announces "NVIDIA drivers are in-use for this build" and `PRE_BUILD` really does rebuild the NVIDIA modules, so the build looks correct end to end. The module is installed as `datadev-gpu-dkms`, and since both variants are `datadev.ko` with the same module name, `lsmod` and `modinfo` cannot tell them apart. The only symptom is `GPUAsync Support : Disabled` in `/proc/datadev_*`, which presents as `GPU Async En : 0` on every card — i.e. it reads as a firmware fault and is not one.

Found on `drp-srcf-gpu006`, where three cards with known-good GPU firmware all reported no GPU capability. Building by hand with `comp_and_load_drivers.sh` was never affected, because it passes `NVIDIA_DRIVERS=<path>` on the make command line, which propagates to the sub-make through `MAKEFLAGS`.

The comment on the `-include` of `version.mk`, immediately below the broken line, already documents this exact trap. The fix is to qualify `Makefile.local` the same way.

Commits:

1. **Refuse to build `datadev-gpu` without NVIDIA** — `build-nvidia.sh` silently    skipped the NVIDIA rebuild and left `NVIDIA_DRIVERS` unset. That is right for CI,    where NVIDIA is absent by design, so it becomes opt-in via `ALLOW_NO_NVIDIA=1` rather than the default. Also ignore the generated `Makefile.local`, which otherwise persists in a working tree and flips a later build silently.
2. **Add a modprobe sample for GPU DAQ nodes** — `datadev.conf` suits CPU-only    machines and is unchanged; GPU nodes get `datadev-gpu.conf` `cfgDevName=1`, `cfgMode=2`, `cfgCont=0`, and small host-side buffers, since the GPU DRP's buffers come from `gpuAddNvidiaMemory()` and are not bounded by `cfgSize`).
3. **Fix `Makefile.local` never reaching the kbuild pass** — the root cause above.
4. **Verify at build time that GPU support was built in** — a `POST_BUILD` hook greps the built module for the `#ifdef DATA_GPU` branch, so a mislabelled package fails the build instead of shipping. `PRE_BUILD` checks the input; this checks the output. The two are different questions and only the first was being asked.
5. **Add `dkms-reload.sh`** — one command from `git pull` to a running, verified driver.

Verified on `drp-srcf-gpu006` and `drp-srcf-gpu001` (Rocky 9, dkms 3.2.2, nvidia 595.91.07). Before: `GPUAsync Support : Disabled` and `GPU Async En : 0` on every card. After: `Enabled` and `1`, from a modprobe-loaded, dkms-installed module, with `/sys/module/datadev/srcversion` matching `modinfo -F srcversion datadev`.

## Release note

Added, in your words, because someone will file that ticket:

> `dkms autoinstall` will now fail loudly for `datadev-gpu-dkms` on a kernel upgrade
> where the NVIDIA tree is not ready, instead of quietly installing a module without
> GPU support.


data_dev/driver/datadev-gpu.conf
```diff
@@ -0,0 +1,35 @@
...
+# does not bound them.  Both can be in use at once on one card: the GPU DRP takes
+# lane 0 and CPU DRPs any other lane, e.g. an ePixUHR3x2 GPU DRP on lane 0 at
+# ~200 kB per DMA alongside a timing CPU DRP on lane 1 that is content with 4 kB.
+#
+#   cfgDevName=1   Name devices by PCI bus number, /dev/datadev_a1 rather than
+#                  /dev/datadev_1, so a name cannot shift when a card is added,
+#                  removed or re-enumerated.  Default 0.
+#   cfgMode=2      BUFF_STREAM: kmalloc + dma_map_single with explicit cache
+#                  synchronisation.  Default 1, BUFF_COHERENT.
+#   cfgCont=0      Continuation off, so a frame larger than the buffer asserts
```
ruck314: This description is correct and the docs contradict it. `cfgCont` only ever reaches `writel(dev->cfgCont ? 1 : 0, &reg->contEnable)` (`common/driver/axis_gen2.c:535`), and the hardware continue bit comes back per descriptor as `flags[16]` and the `Continue Count` in `/proc` (`axis_gen2.c:296`, `:905`). Nothing ties it to buffer return.

But `docs/reference/module-parameters.rst:52` says `1` = driver auto-returns buffers to the free pool and `0` = application must call `dmaRetIndex()`, and `docs/tutorials/first-transfer.rst:93` repeats it. Please fix both here. Leaving two incompatible descriptions of one parameter in the tree is the same failure class this PR exists to remove.

data_dev/driver/datadev-gpu.conf
```diff
@@ -0,0 +1,35 @@
...
+#                  the overflow bit in DmaDsc and the remainder is dropped
+#                  (AxiStreamDmaV2Write.vhd:325), instead of spanning buffers.
```
ruck314: Line number in another repo will rot. Name the signal or register instead of `AxiStreamDmaV2Write.vhd:325`.

data_dev/driver/check-gpu-build.sh
```diff
@@ -0,0 +1,67 @@
...
+set -eu
+
+MODULE=${1:-datadev.ko}
+
+# Makefile.local is how PRE_BUILD reports what it decided: NVIDIA_DRIVERS=<path>
+# when a matching NVIDIA tree was found, and empty when it was told to carry on
+# without one (ALLOW_NO_NVIDIA=1, as CI does).  Taking the intent from there
+# rather than from the environment keeps this consistent with whatever PRE_BUILD
+# actually did, however dkms was invoked.
+if [ ! -s Makefile.local ]; then
```
ruck314, line 39: Correct today, but worth stating the invariant it rests on. dkms runs CLEAN *after* PRE_BUILD (`build_module` calls `prepare_build`, which ends with PRE_BUILD, then `do_build`, whose first step is CLEAN). `make clean` currently leaves `Makefile.local` alone, so PRE_BUILD's output survives into the build.

If anyone later adds `rm -f Makefile.local` to the clean target, which now looks like an obvious tidy-up given it is gitignored, that file is deleted before MAKE runs and this guard takes the skip branch instead of firing: GPU support vanishes and the check passes. One comment on the clean target would close it off.

data_dev/driver/dkms-reload.sh
```diff
@@ -0,0 +1,139 @@
...
+else
+    dkms ldtarball --archive="$TARBALL"
+fi
+dkms build   -m "$PKG" -v "$DVER"
+dkms install -m "$PKG" -v "$DVER" --force
+
+# Retire every other datadev package and version.  They all install the same
+# datadev.ko to the same place, so leaving one behind means modprobe can pick up
+# a stale module -- including one of the other variant.
+dkms status | awk -F'[/,]' '/^datadev(-gpu)?-dkms\//{ print $1"/"$2 }' | sort -u |
```
ruck314: `-F'[/,]'` mangles `added`-state entries. `dkms status` prints those as `datadev-dkms/7.5.0: added`, so `$2` becomes `7.5.0: added`, and `dkms remove -v '7.5.0: added'` fails and is swallowed by the `|| true`. `built`/`installed` lines parse fine, so the practical effect is that a stale added-only registration is never retired. `-F'[/,:]'` fixes it.

data_dev/driver/dkms-reload.sh
```diff
@@ -0,0 +1,139 @@
...
+    echo "       that reloaded it.  Candidates:"                                     >&2
+    echo "         systemctl list-units '*datadev*'"                                 >&2
+    echo "         grep -rl datadev /etc/rc.d/rc.local /etc/rc.local \\"             >&2
+    echo "              /etc/systemd/system /etc/modules-load.d 2>/dev/null"         >&2
+    exit 1
+fi
+echo "    module matches what the next boot will load"
+
+if [ "$VARIANT" = gpu ]; then
+    if grep -h "GPUAsync Support" /proc/datadev_* | grep -qv Enabled; then
```
ruck314, line 137: False pass when the node has no cards: the glob does not expand, grep prints `No such file or directory`, the `if` is false, and the script reports `GPUAsync support enabled on all cards`. Verified. Worth checking that at least one `/proc/datadev_*` exists first, since this block is the verification step.

data_dev/driver/dkms-reload.sh
```diff
@@ -0,0 +1,139 @@
...
+# checkout full of root-owned objects that the owner then cannot clean.
+BUILDER=${SUDO_USER:-root}
+echo "==> make dkms (as $BUILDER)"
+if [ "$BUILDER" != root ]; then
+    sudo -u "$BUILDER" make dkms
+else
+    make dkms
+fi
+
+TARBALL=$(ls -t "$PKG"-*.tar.gz 2>/dev/null | head -1)
```
ruck314: Under `set -euo pipefail` a failing `ls` aborts the script here with status 2 and no output at all, since stderr is discarded, so the error message below is unreachable. `|| true` on the assignment makes the `-z` check do its job.

data_dev/driver/Makefile
```diff
@@ -150,6 +160,7 @@ endef
 	cp -rLv src/* .dkms-common/src/
 	cp -L Makefile .dkms-common/
 	cp -L datadev.conf .dkms-common/
+	cp -L datadev-gpu.conf .dkms-common/
```
ruck314: `.dkms-common` is shared, so the CPU tarball now ships `datadev-gpu.conf` too (confirmed in the built tarball). The `$(4)`/`$(5)` cp lines already take a space-separated list, so it could go there and stay GPU-only.

ruck314 (commented):
Approve on the substance. I reproduced the two-pass include failure against a real kernel build tree: pass 1 emits `-DDATA_GPU=1`, the kbuild pass does not, and `src/gpu_async.o` silently drops out of `$(NAME)-objs` as well. `check-gpu-build.sh` catches exactly that case (passes on a GPU build, fails on a non-GPU build with a non-empty `Makefile.local`, skips on an empty one, survives `strip -g`, handles `.xz`). Checked against dkms source: POST_BUILD runs from the build dir *after* the compression step, so the compressed-module handling is not speculative, and `POST_BUILD` is in `dkms_conf_variables` in 2.8.7, so the ubuntu:22.04 smoke test will not reject `dkms-gpu.conf`. CI is unaffected: the GPU smoke test never triggers PRE_BUILD, and `build-gpu.sh` passes `NVIDIA_DRIVERS=` on the command line.

Inline comments are all minor except the `cfgCont` one, which wants a docs fix in this PR.

Three items with no diff line to attach to:

- `dkms-reload.sh`, `ALLOW_NO_NVIDIA`, and `datadev-gpu.conf` are undocumented. `docs/how-to/build-and-load.rst:60` already points at `datadev.conf` and `data_dev/driver/README.md:75` already documents `comp_and_load_drivers.sh`, so both have an obvious slot.
- Title is a branch name, and it describes commit 1 rather than the headline fix. Suggest `fix(data_dev): make Makefile.local reach the kbuild pass`.
- `origin/pr-check-for-datagpu-en` is already an ancestor of `pre-release` and only touches `common/driver`, so the overlap note at the end of the description can go.

Worth calling out in the release notes: `dkms autoinstall` will now fail loudly for `datadev-gpu-dkms` on a kernel upgrade where the NVIDIA tree is not ready, instead of quietly installing a non-GPU module. That is the intent, and `BUILD_DEPENDS="nvidia"` already orders it on dkms 3.x, but someone will file a ticket.

data_dev/driver/datadev-gpu.conf
rclaus2: Fixed in both places. `docs/reference/module-parameters.rst` now describes what it does — written to the write engine's `contEnable`, so `1` lets a frame larger than `cfgSize` continue into further buffers with `cont` set on all but the last descriptor (flags bit 16, counted as `Continue Count` in `/proc/datadev_*`), and `0` makes the engine assert `overflow` and drop the remainder. It also says explicitly that it has no bearing on when buffers are returned to the free pool, so the old claim does not creep back in.

`docs/tutorials/first-transfer.rst:93` reworded to match.

Worth saying that this is how the parameter came to be set at all: the value we want for GPU nodes is `cfgCont=0`, and it was chosen on the strength of the VHDL, not the docs. Had we trusted `module-parameters.rst` we would have read it as a buffer-return setting and left it alone.

data_dev/driver/datadev-gpu.conf
rclaus2: Agreed. Now names `contEnable`, `overflow`, `dropEn` and `cont` instead of citing `AxiStreamDmaV2Write.vhd:325`.

data_dev/driver/check-gpu-build.sh
rclaus2: Good catch, and the trap is exactly as described: the file is gitignored, which makes `rm -f Makefile.local` in `clean` look like tidying up, and DKMS runs CLEAN after PRE_BUILD, so the effect would be a build that silently loses `DATA_GPU` *and* a guard that passes because it takes the skip branch. Failing open, which is the one thing this check must not do.

Comment added to the `clean` target rather than to the guard, since that is where someone would be standing when they made the change.

data_dev/driver/dkms-reload.sh
rclaus2: Fixed to `-F'[/,:]'`, with a comment recording why `:` is in there, since it looks redundant next to `/` and `,`. You are right that the `|| true` was hiding it — the retire step would silently skip a stale added-only registration forever.

data_dev/driver/dkms-reload.sh
rclaus2: Fixed with a `compgen -G "/proc/datadev_*"` check before the GPUAsync test. This is the verification block, so a false pass there is the worst place for one.

data_dev/driver/dkms-reload.sh
rclaus2: Fixed with `|| true` on the assignment. The `-z` test and its message were unreachable, and with stderr discarded the script would have exited 2 saying nothing at all.

data_dev/driver/Makefile
rclaus2: Moved out of `.dkms-common` into the GPU variant's `$(5)` list, which does take a list. Verified in both built tarballs: CPU now carries `datadev.conf`, `datadev_load.conf` and `dkms.conf` only; GPU carries both samples plus `build-nvidia.sh` and `check-gpu-build.sh`.

rclaus2:
# Replies to review of PR #319 — all fixed in 5946aa40

## Undocumented: dkms-reload.sh, ALLOW_NO_NVIDIA, datadev-gpu.conf

Both slots used, as suggested. `docs/how-to/build-and-load.rst` gains the GPU sample next to the existing `datadev.conf` instructions, plus an "Install and Reload via DKMS" section. `data_dev/driver/README.md` gains the same, and now also says that `comp_and_load_drivers.sh` installs nowhere `modprobe` can find and takes its parameters from its own command line rather than `/etc/modprobe.d` — which is what made the stale install on our node so hard to see.

Both mention that `ALLOW_NO_NVIDIA=1` exists for CI and produces a driver with no GPU support.

## Title

Taking your suggestion: `fix(data_dev): make Makefile.local reach the kbuild pass`. Mine was a branch name describing commit 1 rather than the headline fix.

## The pr-check-for-datagpu-en note

Confirmed and removed — `git merge-base --is-ancestor origin/pr-check-for-datagpu-en origin/pre-release` returns true, so it is already in.

## Release note

Added, in your words, because someone will file that ticket:

> `dkms autoinstall` will now fail loudly for `datadev-gpu-dkms` on a kernel upgrade
> where the NVIDIA tree is not ready, instead of quietly installing a module without
> GPU support.

## On the reproduction

Thank you for reproducing the two-pass failure against a real kernel tree, and particularly for noticing that `src/gpu_async.o` drops out of `$(NAME)-objs` as well — we had only tracked the missing `-DDATA_GPU=1`. Also for checking POST_BUILD against the DKMS source: that it runs from the build dir after compression, and that it is in `dkms_conf_variables` in 2.8.7, are both things we assumed from observed behaviour on Rocky 9 with dkms 3.2.2 rather than verified across versions.

data_dev/driver/check-gpu-build.sh
ruck314, line 39: `-s` is false for an absent file as well as an empty one, so if `Makefile.local` is missing here (PRE_BUILD did not run, or a future dkms runs POST_BUILD from another cwd) the check skips and passes. Only the empty file means "built without NVIDIA by request". Suggest failing closed on absent, as the missing-module case below already does:

```sh
if [ ! -e Makefile.local ]; then
    echo "ERROR: no Makefile.local in $PWD; PRE_BUILD did not run here." >&2
    exit 1
fi
if [ ! -s Makefile.local ]; then
```

data_dev/driver/build-nvidia.sh
```diff
@@ -31,13 +31,34 @@ fi
...
+        echo "ERROR: no NVIDIA DKMS module found for kernel $VERSION."       >&2
+        echo "       Looked for: dkms status nvidia -k $VERSION"             >&2
+        echo "                   $SRCTREE/<nvidia-version>/"                 >&2
+        echo "       Continuing would build datadev-gpu WITHOUT DATA_GPU,"   >&2
+        echo "       producing a module that cannot do GPU DMA while still"  >&2
+        echo "       being named datadev-gpu.  Install and register the"     >&2
+        echo "       NVIDIA open kernel modules for this kernel first."      >&2
+        echo "       To build without GPU support anyway, as CI does,"       >&2
+        echo "       set ALLOW_NO_NVIDIA=1."                                 >&2
+        exit 1
```
ruck314, line 57: This breaks `scripts/ci/test-dkms-gpu.sh`: it runs `dkms install`, which triggers this PRE_BUILD, on a cell with no NVIDIA DKMS tree (`build-gpu.sh` points `NVIDIA_DRIVERS` at `emulator/gpu_stub`; nothing is registered with dkms). `scripts/ci-local/run_cell.sh --phase gpu --load-test 1` calls it. GitHub CI is unaffected, it only runs the ldtarball smoke test. `run_cell.sh` also does not propagate that script's exit status, so the cell still passes and only the log shows the failure. `export ALLOW_NO_NVIDIA=1` in `test-dkms-gpu.sh`, with a line on why, fixes it.

ruck314 (commented):
Two follow-ups from a second pass, both small. I would also like Jeremy to review and approve this before the merge

data_dev/driver/check-gpu-build.sh
rclaus2: Right, and thank you: that is the guard failing open, which is the single outcome it exists to prevent. Fixed as you suggested, with the distinction written down so the two cases do not get merged again — only an empty `Makefile.local` means "built without NVIDIA by request"; absent means there is nothing to go on.

Verified all four branches: absent → 1, empty → 0 and skip, non-GPU module with a populated `Makefile.local` → 1, GPU module → 0.

Worth noting this is the second time the same mistake nearly landed in this file. The first was looking only for `datadev.ko` when dkms had already compressed it, which also produced a pass-shaped result from a check that could not actually see its subject.

data_dev/driver/build-nvidia.sh
rclaus2: Confirmed, and the diagnosis is exact. `build-gpu.sh:67` passes `NVIDIA_DRIVERS=$(pwd)/emulator/gpu_stub` on the make command line, which dkms cannot see, so the `dkms install` at `test-dkms-gpu.sh:80` triggers `PRE_BUILD` with nothing registered and the new refusal fires. `export ALLOW_NO_NVIDIA=1` added there, with a comment on why and on the consequence: the module that cell installs has no GPU support, by design, since the test is about the tarball and install path.

My earlier claim that CI was unaffected was wrong in a way worth admitting: I checked `.github/workflows` and the `build-{deb,rpm}.sh` scripts and stopped, so I never looked at `scripts/ci-local/` at all. GitHub Actions was indeed unaffected; the local harness was not.

rclaus2:
### Two observations, neither for this PR

**`run_cell.sh` discards `test-dkms-gpu.sh`'s status.** At lines 262-263:

```sh
[ $rc -eq 0 ] && bash scripts/ci/test-dkms-gpu.sh; \
exit $rc;
```

`rc` was captured from `test-gpu.sh` before `test-dkms-gpu.sh` ran, and `exit $rc` uses that stale value. So any failure in the dkms test is invisible to the cell — which is precisely why the breakage above would not have failed anything, as you say. Ours is just the first change to trip it. Left alone as out of scope, but a test whose result cannot fail is worth someone's attention.

**The GPU dkms cell no longer verifies GPU support.** With `ALLOW_NO_NVIDIA=1` the `POST_BUILD` check skips, so `--load-test 1` installs a `datadev-gpu`-named package without GPU support — the state this PR exists to make impossible on a real node. Acceptable in a packaging test, and the alternative would be registering the stub with dkms so `PRE_BUILD` finds it and the check stays live. That is a larger change than this PR should carry, but it would close the gap if you think the cell should cover it.

JJL772 (approved):

PR #321, opened 2026-09-14
Enforce GPU memory alignment

JJL772:
### Description

Enforces GPU memory alignment. Since we are aligning the virtual memory address anyways, we should enforce this requirement on incoming memory. If userspace supplies us a block of memory that isn't aligned to the GPU page size, we mask off that offset and write to a location that userspace doesn't expect, resulting in an indefinite hang (the handshake location never gets set).

This should be merged after #319 and should be a new minor version of the driver.


data_dev/app/src/GpuAsyncLib.cpp
```diff
@@ -280,3 +280,73 @@ void gpuUnmapFpgaMem(GpuDmaBuffer_t* mem) {
...
+        return res;
+    }
+
+    return res;
+}
+
+void vmmCuFree(CudaVMMAlloc& alloc) {
+    if (alloc.ptr == 0)
+        return;  // Nothing to free!
+    cuMemAddressFree(alloc.ptr, alloc.size);
```
ruck314, line 350: Blocking: `vmmCuFree` releases the handle and the address reservation but never unmaps the range, so the physical GPU memory is leaked.

The CUDA VMM API keeps the backing allocation alive as long as either a mapping or an unreleased handle references it. `cuMemRelease` only drops the handle reference; the mapping created by `cuMemMap` in `vmmCuAlloc` still holds one. The driver API docs for `cuMemUnmap` state that "backing memory allocations will be freed if there are no existing mappings and there are no unreleased memory handles", so without the unmap this memory is only reclaimed when the context is destroyed. Freeing the address reservation while it still has a live mapping is also outside what the API documents as valid.

For `rdmaTest` the process exits right after, so the leak is invisible. But this is a public helper in `GpuAsyncLib.h`, and any longer-lived consumer that allocates and frees buffers repeatedly will exhaust device memory.

Suggested teardown order, matching the error path already used at line 336:

```cpp
cuMemUnmap(alloc.ptr, alloc.size);
cuMemAddressFree(alloc.ptr, alloc.size);
cuMemRelease(alloc.handle);
```

data_dev/app/src/rdmaTest.cu
```diff
@@ -216,30 +247,22 @@ static int initSession(TestSession& s, const char* dev, int gpuIdx,
 
     s.dmaHeaderSize = s.coreRegs->dmaDataBytes();
 
+    const size_t totalBufSize = ALIGN_VALUE(bufSize + s.dmaHeaderSize, GPU_PAGE_SIZE);
+
     /* Allocate rx (and optionally tx) buffers sized bufSize + dmaHeaderSize for
      * descriptor headroom; only bufSize is registered with the FPGA. */
```
ruck314: Stale comment. Before this PR the buffers were `cudaMalloc`'d at `bufSize + dmaHeaderSize` but only `bufSize` was passed to `gpuAddNvidiaMemory`, which is what this sentence describes. Now `allocGpuMem` passes `totalBufSize` (the 64k-aligned `bufSize + dmaHeaderSize`) through to `gpuAddNvidiaMemory`, so the full allocation is registered with the FPGA and drives `RemoteWriteMaxSize`.

That behavior change is fine, and arguably safer since the FPGA writes header plus payload into this buffer, but the comment now says the opposite of what the code does. Worth rewording so the next reader does not go looking for a `bufSize`-only registration that no longer exists.

data_dev/app/src/rdmaTest.cu
```diff
@@ -51,6 +49,11 @@ static int str2int(const char* s) {
     return strtol(s, NULL, base);
 }
 
+/* CUDA Allocations must be aligned to this size */
+#define GPU_PAGE_SIZE 0x10000
+
+#define ALIGN_VALUE(_x, _align) (((_x) + (_align) - 1) & ~((_align) - 1))
```
ruck314, line 55: Non-blocking: this PR adds `alignValue()` to `GpuAsyncLib.h`, and this file already includes that header, so `ALIGN_VALUE` is a second implementation of the same thing living a few hundred lines from the first. `GPU_PAGE_SIZE` is also a third copy of the 64k constant (the driver has `GPU_BOUND_SIZE`, and line 171 hardcodes `0x10000` for the size check).

Two alignment helpers with different names and forms (macro vs. template) is the kind of thing that drifts apart later. Suggest using `alignValue(bufSize + s.dmaHeaderSize, GPU_PAGE_SIZE)` here and dropping the macro. Consolidating the page-size constant would be nice too but is a larger change and fine to leave.

data_dev/app/src/rdmaTest.cu
```diff
@@ -155,6 +184,8 @@ static int initSession(TestSession& s, const char* dev, int gpuIdx,
         return -1;
     }
 
+    cuCtxSetCurrent(s.cuda.context());
```
ruck314: Non-blocking: this call is redundant and its return value is discarded.

`s.cuda.init(gpuIdx)` on line 182 ends with `cuCtxCreate`, and per the driver API that pushes the new context onto the calling thread's stack, making it current. Since `initSession` runs on the same thread that just created it, the context is already current when this line executes.

If there is a real case where the context is not current here (for example a future refactor that creates it on another thread), then the call is doing load-bearing work and its `CUresult` should be checked like the other driver calls in this file. Either drop it or wrap it in `assertOk`; the silent unchecked form is the one option that hides a failure.

common/driver/gpu_async.c
```diff
@@ -200,6 +200,13 @@ int32_t Gpu_AddNvidia(struct DmaDevice *dev, uint64_t arg) {
       return -EINVAL;
    }
 
+   // Memory must be aligned to GPU page boundary to avoid GpuAsyncCore writing out-of-bounds
+   if ((dat.address & GPU_BOUND_MASK) != dat.address) {
```
ruck314, line 206: The check itself is correct and I agree with enforcing it. One follow-on: this makes the unaligned-address handling below it dead code.

Once `dat.address` is required to be 64k aligned, `virt_offset` on line 248 is always zero and `dma_address = dma_addresses[0] + virt_offset` on line 283 reduces to `dma_addresses[0]`. The comment on lines 246-247 ("Handle addresses that aren't aligned to 64k boundary. CUDA doesn't have an easy way to perform aligned allocations, so account for that here") now describes a path that cannot be reached, and its premise is exactly what this PR fixes in userspace with `vmmCuAlloc`.

Leaving it as-is is not a functional problem, but a reader of the driver will see contradictory intent within forty lines: one block rejects unaligned addresses, the next explains how it handles them. Suggest removing `virt_offset` and the comment in this PR since this is the change that obsoleted them.

ruck314 (changes requested):
Driver check looks right. One blocking item (leak in vmmCuFree), the rest are cleanup and optional. I pushed a whitespace-only commit to get the cpplint step green.

common/driver/gpu_async.c
JJL772: I have removed the `virt_offset` handling

ruck314 (approved):
@JJL772 LGTM.  There's one more unresolved inline comment left (non-blocking) if you want to address it or close it.  Merge when you are ready

PR #322, opened 2026-09-16
include/GpuAsyncLib: Make alignValue compatible with different base types

JJL772:
### Description

Fixes build error and makes alignValue a bit easier to use.


ruck314 (approved):
@JJL772 LGTM. Merge when ready

PR #323, opened 2026-09-18
PR verify driver reload and rdmatest timeout

rclaus2:
Report when a reload or a wait did not do what it appeared to.

## Description
Two independent fixes, both of the same kind: a step that reported success, or nothing at all, without testing the thing that mattered.

### `dkms-reload.sh`: verify the loaded module's version, not its source hash

The script reported a successful reload on drp-srcf-gpu001 while the module loaded four days earlier was still resident and the newly built one sat unused on disk. It printed `module matches what the next boot will load`, then `done`, with no error.

The verification compared `srcversion`:

```sh
RUNNING=$(cat /sys/module/datadev/srcversion)
ONDISK=$(modinfo -F srcversion datadev)
```

`srcversion` is a hash of the `.c`/`.h` files. #319 changed only the Makefile, the build scripts and the docs, so it was byte-identical between the two builds and the comparison passed whichever module was resident. The message was even honest — "what the next boot will load" is a statement about `modinfo`, i.e. about disk — and was read as though it said "what is running".

`GITV` is compiled in per build and printed by `dma_common.c` as `DMA Driver's Git Version` in `/proc/datadev_*`, so it is the field that discriminates. The check now compares that against the version just installed, and fails with the remedy and the commands to find what holds the module. The `srcversion` comparison is kept as a secondary test: it is still meaningful when sources do change, and it catches a module loaded from another path.

Tested against the broken state — the new check errors where the old one passed — and against a genuine conversion on drp-srcf-gpu008, `7.5.1` to `7.6.0-27`, where version and sources both differ and both checks pass for real reasons.

### `rdmaTest`: say when nothing is arriving, instead of hanging silently

With no transmitter running, `rdmaTest` blocked forever with no output. That is the normal first-time experience rather than an edge case, because the program starts no data source: InterCardTest holds its PRBS generator off until `PrbsTx.TxEn` is set from `interCardGui.py`, and nothing in the program or its `--help` says so. A hang with no output sends the reader to the source code to find that out.

It now prints one line naming what to check, then a dot a second, and keeps waiting — it does not give up, because the useful order of work is to start the program and *then* the transmitter. `-T` sets the delay before it speaks (default 10 s); `-T 0` restores the old silence.

**The ordering is the whole fix, and the hang is not where it appears to be.** A `gstack` showed the blocked thread inside the *enqueue* of `cuMemcpyDtoHAsync`, not in `cuStreamWaitValue32`:

```
#10 cuMemcpyDtoHAsync_v2 ()
#11 runSimpleLoop (s=...)
```

`hdr` is an ordinary local, so the destination is pageable host memory, and CUDA documents such device-to-host copies as behaving synchronously — staging through an internal pinned buffer and waiting for the stream. Enqueued while the stream is still blocked on the doorbell, it blocks the host before any polling code is reached. So the wait is now drained *before* the copy is enqueued; the stream is then idle and the copy returns promptly.

The fast path does not pay for the polling. Sleeping between attempts would add the sleep to every event's latency and cap the test near 1 kHz, which would make the GiB-transferred figure the tool already prints meaningless. It spins for the first millisecond — far longer than a wait takes when data is flowing — and only then sleeps.

Tested on drp-srcf-gpu008 both ways: against a card with no source it warns and dots indefinitely and exits cleanly on `^C`; against a card running InterCardTest firmware with `PrbsTx.TxEn` set it sustains 65536-event reporting intervals every few seconds, so the polling costs nothing measurable.

### Not addressed here

Noticed while working on this, left alone as out of scope:

- `emulator/gpu_stub`'s `clean` target recurses into `$(KERNELDIR)`, so `make clean` fails when the kernel headers are absent — which is what broke `GPU Test (rockylinux:9)` on `pre-release` for a while. `data_dev/driver/Makefile` has the same unguarded pattern.
- Nothing in CI builds `make cuda`, which is how #321 merged with `rdmaTest.cu` uncompilable.
- `rdmaTest`'s summary prints `%-4lu` for the invalid-event count, whose padding reads as a missing number, and labels a decimal `1.0E+9` divisor as `GiB`.
- Every event pays a synchronous pageable device-to-host copy for a 16-byte header. Pinning it with `cuMemAllocHost` would make it genuinely asynchronous.
- `software/scripts/interCardGui.py` in `axi-pcie-devel` does `int(args.dev.split('_')[1])`, which rejects the hex device names `options datadev cfgDevName=1` produces. Dropping the `int()` is sufficient, since `PcieControl` interpolates the value straight into a path.  Other TID tools likely have the same assumption.


data_dev/app/src/rdmaTest.cu
```diff
@@ -353,11 +414,19 @@ static void runSimpleLoop(TestSession& s) {
...
             s.rxBuffers[curBuff].ptr + 4, 1, CU_STREAM_WAIT_VALUE_GEQ));
 
-        /* Download header data immediately. */
+        /* Wait for the doorbell *before* enqueueing the copy below.  cuMemcpyDtoHAsync
+         * into pageable memory -- hdr is an ordinary local -- is documented as behaving
+         * synchronously: the driver stages through an internal pinned buffer and waits for
+         * the stream.  Enqueued while the stream is blocked on the wait above, it therefore
+         * blocks the host inside the enqueue, which is where an untimed run actually hangs.
+         * Draining the wait first leaves the stream idle, so the copy does not block. */
+        syncWaitingFor(s, "event from the FPGA");
```
JJL772, line 423: Adding a stream sync here will artificially slow down the FPGA's latency monitoring, since it's basically waiting for buffer 0 to be returned to the FPGA. This code should be wrapped in `if (timeout > 0) {...}` (i.e. "no timeout, wait forever") to avoid this disruption.

data_dev/app/src/rdmaTest.cu
```diff
@@ -375,9 +444,11 @@ static void runSimpleLoop(TestSession& s) {
                         hdr.size, maxPayload);
                 invalidEvents++;
             } else {
-                /* Wait for the FPGA to return the free list back to GPU. */
+                /* Wait for the FPGA to return the free list back to GPU.  Drained here
+                 * for the same reason as above, before anything else is enqueued. */
                 assertOk(cuStreamWaitValue32(s.stream,
                     s.txBuffers[curBuff].ptr, 1, CU_STREAM_WAIT_VALUE_GEQ));
+                syncWaitingFor(s, "buffer returned by the FPGA");
```
JJL772: Wrap this call in `if (timeout > 0) { ... }`

data_dev/app/src/rdmaTest.cu
```diff
@@ -335,6 +339,63 @@ static void cleanupSession(TestSession& s) {
...
+ * It reports and keeps waiting rather than exiting, because the useful order of work is to
+ * start this program and then the transmitter.  One line, then a dot a second.
+ *
+ * The fast path must not pay for the polling: sleeping between attempts would add the sleep
+ * to every event's latency and cap the test near 1 kHz, so it spins for the first
+ * millisecond -- far longer than a wait takes when data flows -- and only then sleeps.
+ */
+static void syncWaitingFor(TestSession& s, const char* what) {
+    if (s_warnSecs <= 0.0) {                /* Reporting disabled: plain blocking sync */
+        assertOk(cuStreamSynchronize(s.stream));
```
JJL772, line 360: As noted in my comments below, I think we want to avoid another stream sync here.

data_dev/app/src/rdmaTest.cu
rclaus2: ## I think this one is a misreading, please check my reasoning

`syncWaitingFor()` **replaces** the `cuStreamSynchronize()` that was already on that line; it does not add one. Before:

```c
assertOk(cuStreamWaitValue32(s.stream, s.rxBuffers[curBuff].ptr + 4, 1, GEQ));
assertOk(cuMemcpyDtoHAsync(&hdr, s.rxBuffers[curBuff].ptr, sizeof(hdr), s.stream));
assertOk(cuStreamSynchronize(s.stream));        // <- blocks here until the event arrives
```

After:

```c
assertOk(cuStreamWaitValue32(s.stream, s.rxBuffers[curBuff].ptr + 4, 1, GEQ));
syncWaitingFor(s, "event from the FPGA");       // <- blocks here instead, and can speak
assertOk(cuMemcpyDtoHAsync(&hdr, s.rxBuffers[curBuff].ptr, sizeof(hdr), s.stream));
assertOk(cuStreamSynchronize(s.stream));        // <- now returns promptly, stream is idle
```

So the count of synchronization points per iteration is unchanged. What moved is *which* one blocks: the wait is drained before the copy is enqueued rather than after.

That reordering is the substance of the change, and it is not cosmetic. The hang in an untimed run is not in `cuStreamWaitValue32` and not in `cuStreamSynchronize` — it is inside the **enqueue** of `cuMemcpyDtoHAsync`. `hdr` is an ordinary local, so the destination is pageable host memory, and CUDA documents device-to-host copies into pageable memory as behaving synchronously: the driver stages through an internal pinned buffer and waits for the stream. Enqueued while the stream is still blocked on the doorbell, it blocks the host before any polling code can run. A `gstack` on the hung process:

```
#10 cuMemcpyDtoHAsync_v2 ()
#11 runSimpleLoop (s=...)
```

I tried wrapping the timeout around the existing sync first, exactly as suggested, and it produced a process that spun silently with no output at all — which is how I found the enqueue was the blocking call.

On the polling cost, since that is the reasonable worry: `syncWaitingFor()` spins on `cuStreamQuery` for the first millisecond before it starts sleeping, so the steady-state path does not pay a sleep per event. Measured on an H200 against InterCardTest firmware with the PRBS generator running, it sustains 65536-event reporting intervals every few seconds — tens of thousands of events per second. A 1 ms sleep per poll would have capped it near 1 kHz, which is why the spin is there.

If you would still rather the receive site were unconditional when `-T 0` is given, that already happens: `syncWaitingFor()` short-circuits to a plain `cuStreamSynchronize()` when `s_warnSecs <= 0`, so `-T 0` restores the original behavior exactly. What it cannot do is skip the *reordering*, since that is what stops the hang.

data_dev/app/src/rdmaTest.cu
rclaus2: You are right. That `syncWaitingFor()` is a new synchronization point I added, in the loopback path where the code previously only enqueued `cuStreamWaitValue32` and carried on. Waiting there for buffer 0 to come back does hold up the pipeline for no benefit, and the timeout serves no purpose at that site: if the FPGA is not returning buffers, the *receive* wait above will already have said so.

Removing it entirely rather than wrapping it. One bounded wait per iteration is enough, and it is on the path that actually stalls when no data arrives.

data_dev/app/src/rdmaTest.cu
rclaus2: Same comment as for `rdmaTest.cu:423`.

rclaus2:
## What I will push

- Remove the `syncWaitingFor()` call at the loopback site, restoring the plain   `cuStreamWaitValue32` and the single `cuStreamSynchronize` that follows the copies.
- Leave the receive site as it is, unless the reasoning above does not convince you.
- Re-test both paths: no-source (warns and dots) and with-source (rate unchanged).

rclaus2:
# Correction to post on PR #323

Correcting myself hopefully before you find it: I said "the count of synchronization points per iteration is unchanged". That is **not right** for the receive path, and I should have counted rather than asserted.

Accurately:

| path | before | after |
|---|---|---|
| non-loopback | 1 sync | **2** — the bounded wait, plus the sync that follows the copy |
| loopback | 1 sync | 1 — unchanged, and now unchanged by this PR too |

So on the receive path there is one more call than before. What I should have said is that there is no additional *blocking* call: the sequence is now

```c
cuStreamWaitValue32(...)            // enqueue the wait
syncWaitingFor(...)                 // blocks here until the doorbell clears  <-- the wait
cuMemcpyDtoHAsync(&hdr, ...)        // stream is idle, so this does not block
cuStreamSynchronize(...)            // returns promptly, nothing outstanding  <-- added
```

The added `cuStreamSynchronize` sits after the copy, and by then the only thing in the stream is a 16-byte device-to-host copy on an already-drained stream, so it costs a syscall rather than a wait. Whether that is worth it is a fair question and yours to judge — but it is not free, and I claimed it was.

The reason it cannot simply be dropped is that the copy must be complete before `hdr` is read on the next line, and `cuMemcpyDtoHAsync` gives no such guarantee on its own. Upstream relied on the single sync for both purposes: waiting for the event *and* completing the copy. Once those are separated — which is what fixes the hang — they need one each.

If the extra syscall per event matters, the alternative is to pin `hdr` with `cuMemAllocHost`, which would make the copy genuinely asynchronous and let a single sync serve again. That is a larger change and touches the per-event round trip you and I have already agreed is a separate performance question, so I have left it out of this PR.

Sorry for the oversight!

rclaus2:
# A separate finding, worth its own comment on the PR

While testing the loopback path I hit something that is not about this PR but is about the dkms packaging, so it seems worth raising here.

**The dkms path does not carry over the NVIDIA module parameters that `comp_and_load_drivers.sh` passes.** That script loads nvidia with

```
insmod nvidia.ko NVreg_OpenRmEnableUnsupportedGpus=1 NVreg_EnableStreamMemOPs=1
```

but nothing in `dkms.conf`, `build-nvidia.sh` or `dkms-reload.sh` supplies them, and a boot-time `modprobe` has no command line. So converting a node from `comp_and_load_drivers.sh` to dkms silently loses both, and the loss is invisible until something needs them.

**`rdmaTest` is one of the things that needs them.** It does its FPGA handshake from the host with `cuStreamWriteValue32`/`cuStreamWaitValue32`, checks `CU_DEVICE_ATTRIBUTE_CAN_USE_STREAM_MEM_OPS_V1` at startup, and aborts:

```
WARNING: device does not support CUDA Stream Operations; this code may not run.
Selected GPU lacks stream memory ops; aborting
```

Confirmed by reading the attribute directly — 0 on all five GPUs, with `EnableStreamMemOPs: 0` in `/proc/driver/nvidia/params`. The tool had worked on the same node earlier the same day, before a reboot, because nvidia had then been loaded by `comp_and_load_drivers.sh`.

Worth noting `drp_gpu` does *not* need the flag — its kernels write the GpuAsyncCore registers directly through the `CAP_SYS_ADMIN` `IOMEMORY` mapping — so a dkms-converted node runs the DAQ perfectly while `rdmaTest` refuses to start on it. That asymmetry is what makes it easy to miss.

Two things you might consider, neither of which I have put in this PR since they are
packaging rather than code:

1. Ship a `modprobe.d` snippet with the dkms package, or document in the driver README that the insmod parameters are not carried over. Locally I have added:

   ```
   options nvidia NVreg_OpenRmEnableUnsupportedGpus=1 NVreg_EnableStreamMemOPs=1
   ```

   as `/etc/modprobe.d/nvidia-daq.conf` — a distinct name, since the packaged `/usr/lib/modprobe.d/nvidia.conf` already sets three unrelated parameters.

2. Is `NVreg_OpenRmEnableUnsupportedGpus=1` still needed? I have carried it over so a converted node matches the insmod path, but I do not know why it is there and it may be vestigial.

With the parameter set, the non-loopback path runs correctly on gpu008: it printed the waiting line, dotted while I enabled `PrbsTx.TxEn` from `interCardGui.py`, printed `Resumed: got event from the FPGA`, and then reported steadily to 1.18M events with zero invalid events before I interrupted it. So both halves of the timeout behavior are exercised on real hardware.

**Loopback is now tested too**, with the change you asked for in place: 1.44M events, zero invalid, 378 GB transferred. And it ended by demonstrating the feature better than a clean run would have:

```
1441792 events, 0    invalid events, 377.96 GiB transferred
Waiting for event from the FPGA: is a transmitter enabled?  InterCardTest needs PrbsTx.TxEn ...
.........^C
```

I turned `PrbsTx.TxEn` off before interrupting, so that is a deliberate test of the case the timeout is really for: a source that stops *after* a period of apparently normal operation.  Previously that was a silent hang, and harder to diagnose than a hang at startup because nothing distinguishes it from a quiet detector.

data_dev/app/src/rdmaTest.cu
JJL772: Sorry for the late reply.

> syncWaitingFor() replaces the cuStreamSynchronize() that was already on that line; it does not add one. Before:

The diff doesn't include a stream sync on this line before the `syncWaitingFor`. Since `syncWaitingFor` has a stream sync in it, there's two now. Not a huge ideal imo, since we are already syncing to make the header available to the host, however it does add additional latency.

JJL772:
I'm wondering if it'd be possible to have a control flow like this:
1. Spin wait on handshake space for a max of N cycles
  a. N is determined by `frequency * timeout` (we can query frequency from CUDA)
2. Check handshake space + 0x4 for >= 0x1 on the host side
  a. first memcpyDtoH to get the flag to host mem, and do the check after the stream sync
3. If it is 0x0 we know the transfer timed out, so print a message and go back to the `spinWait` for this buffer again

However I think this isn't worth implementing here. The `if (timeout > 0) {...}` guard around the `syncWaitFor` is enough

data_dev/app/src/rdmaTest.cu
```diff
@@ -393,7 +466,8 @@ static void runSimpleLoop(TestSession& s) {
                     s.txBuffers[curBuff].ptr, 0, 0));
 
                 /* SYNC the stream for the data copy and GPU side free list update
-                 * before triggering the FPGA's doorbell. */
+                 * before triggering the FPGA's doorbell.  Unbounded is fine here: the wait
```
ruck314: This comment was accurate in the previous commit and is inverted now. With the drain at the tx wait removed, the `cuStreamWaitValue32` on `txBuffers[curBuff]` is still pending in the stream when this runs, so this `cuStreamSynchronize` does block until the FPGA returns the buffer.

Please reword to say that this sync waits for the free-list return plus the copies, and that a stall here is expected to be caught first by the receive wait above.

data_dev/app/src/rdmaTest.cu
```diff
@@ -335,6 +339,63 @@ static void cleanupSession(TestSession& s) {
...
+                waiting = true;
+            } else {
+                fputc('.', stderr);         /* stderr is unbuffered, so it appears at once */
+            }
+            nextMark = now + std::chrono::seconds(1);
+        }
+
+        /* Spin for the first millisecond so the fast path is unaffected, then back off. */
+        if (now - start > std::chrono::milliseconds(1))
+            std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
ruck314: Once the first millisecond passes, every poll sleeps 1 ms, so an event arriving after that point waits up to a full sleep quantum before the host reacts. The host's reaction is what closes the FPGA's GpuLatency measurement, so at trigger rates below roughly 1 kHz this adds up to a millisecond of jitter to every reported latency. At the tens of kHz you tested it is invisible, but the README promises accurate latency figures without qualifying the rate.

Suggest spinning until `nextMark` is first reached (i.e. until the warning prints) and only then starting to sleep. By that point nothing is arriving, so latency no longer matters, and the fast path is unchanged.

data_dev/app/src/rdmaTest.cu
```diff
@@ -375,7 +444,11 @@ static void runSimpleLoop(TestSession& s) {
                         hdr.size, maxPayload);
                 invalidEvents++;
             } else {
-                /* Wait for the FPGA to return the free list back to GPU. */
+                /* Wait for the FPGA to return the free list back to GPU.  Deliberately
```
ruck314: Nit on the rationale, not the change: "if the FPGA is not returning buffers, the receive wait has already said so" conflates two independent FPGA paths. Data delivery (write engine) and loopback readback (read engine) can fail separately, so a stalled read engine still hangs silently at the sync below. Leaving the drain out is still the right call per JJL772's comment; I would just not claim the receive wait covers this site.

data_dev/app/src/rdmaTest.cu
```diff
@@ -494,6 +568,8 @@ static void showHelp() {
     printf("  -f FILE      : Dump the first event received to this file\n");
     printf("  -x NUM       : Dump the first NUM bytes of the payload to stdout\n");
     printf("  -c CNT       : Number of events to receive before exiting\n");
-    printf("  -l           : Enable loopback mode: FPGA -> GPU -> FPGA transactions\n");
+    printf("  -l           : Enable loopback mode: FPGA -> GPU -> FPGA transactions\n"
+        "  -T SECS      : Warn if no event arrives within SECS, and keep waiting (default 10,\n"
```
ruck314: Nit: give `-T` its own `printf` line like the other options rather than appending it to the `-l` string. A sentence in the Running section of `data_dev/app/README.md` would also help, since that is where the "configure your firmware to generate frames" hint that this warning points at already lives.

data_dev/driver/dkms-reload.sh
```diff
@@ -111,6 +111,37 @@ fi
...
 echo "==> verifying"
+# The resident module's version, not the package's and not the source hash.  srcversion
+# is a hash of the .c/.h files, so any change confined to the Makefile or the build
+# scripts leaves it identical between two builds -- and then comparing it proves nothing
+# about which module is loaded.  That is not hypothetical: it let a run on
+# drp-srcf-gpu001 report success while the module from four days earlier was still
+# resident, with the new one sitting unused on disk.  GITV is compiled in per build
+# (dma_common.c, "DMA Driver's Git Version"), so it is the field that discriminates.
+if compgen -G "/proc/datadev_*" >/dev/null; then
+    LOADED=$(grep -h "DMA Driver's Git Version" /proc/datadev_* |
```
ruck314, line 125: Under `set -euo pipefail` this guard is unreachable. If none of the `/proc/datadev_*` files contain the version line, `grep` exits 1, `pipefail` propagates that through `head`/`sed`/`tr`, and `set -e` exits the script before the `if [ -z "$LOADED" ]` message prints. The script still fails, just silently with status 1. Appending `|| true` to the pipeline makes the message reachable. Non-blocking.

ruck314 (changes requested):
The dkms-reload half looks right to me: GITV is baked in per build and frozen in version.mk, so comparing it against the tarball's PACKAGE_VERSION is the correct discriminator, and I would approve it on its own. Requesting changes for two items in rdmaTest.cu, inline below. I have left out what JJL772 already raised.

data_dev/app/src/rdmaTest.cu
rclaus2: You are right, and I have stopped defending it — the second sync is gone.

You were also right the first time, in a way I did not see: wrapping the existing sync *is* the correct shape. It only failed because of the pageable-memory problem, and the right fix is to remove that rather than to work around it. `hdr` is now allocated once with `cuMemAllocHost`, so the copy is genuinely asynchronous, the enqueue no longer blocks, and the reordering is unnecessary. The loop is back to upstream's ordering:

```c
cuStreamWaitValue32(...)                    // enqueue the wait
cuMemcpyDtoHAsync(s.hdr, ...)               // pinned: enqueue only, does not block
syncWaitingFor(s, "event from the FPGA");   // the single sync, now bounded
```

Synchronizations per iteration over the whole loop body, counted rather than asserted this time:

| version | syncs |
|---|---|
| upstream | 2 |
| my previous two commits | 3 |
| now | **2** |

So this is at parity with upstream, and the only remaining difference on that line is that the sync can say something when it waits too long. It also removes the per-event pageable staging copy of a 16-byte header, which I had listed in the PR description as a known cost to be dealt with separately — your objection is what made it the obvious thing to fix now instead.

Retested on gpu008 against InterCardTest firmware in loopback mode, and the run happened to exercise both halves of the timeout end to end: dots with no transmitter, `Resumed` and 327680 events with zero invalid after `PrbsTx.TxEn` was set, then dots again after clearing it.

One note on the red check, before you wonder: `Phase 2: CPU Test (rockylinux:9)` fails on this head, and we do not believe it is related. It is the `emulator/gpu_stub` `clean` problem already listed under "Not addressed here" in the PR description — `clean` recurses into `$(KERNELDIR)`, and this runner came up with `CI_HOST_MATCH: 0` and no headers for its own `CI_KVER`:

```
make[1]: *** /lib/modules/5.14.0-687.47.1.el9_8.x86_64/build: No such file or directory.  Stop.
make: *** [Makefile:61: clean] Error 2
```

These two commits touch only `rdmaTest.cu` and `dkms-reload.sh`, and this branch has hit the same failure twice before and passed twice since with nothing in that path changing. Happy to be told otherwise if you read it differently.

data_dev/app/src/rdmaTest.cu
rclaus2: Correct on both counts, and thank you for catching it — the comment said the opposite of what the code does. Removing the drain at the tx wait, which you asked for in the previous round, is exactly what made it wrong: the `cuStreamWaitValue32` on `txBuffers[curBuff]` is still pending in the stream when that sync runs, so it does block until the FPGA returns the buffer.

Reworded along the lines you suggested — it now says the sync waits for the free-list return as well as the copies, and that a stall there is expected to be reported first by the receive wait above.

data_dev/driver/dkms-reload.sh
rclaus2: Confirmed and fixed. I reproduced it in isolation rather than reasoning about it: with the pipeline as it was, the guard never printed and the script exited 1 silently; with `|| true` appended the message appears and the exit status is still 1. Exactly as you described.

Marked non-blocking on your side, but it was worth fixing here since the whole point of this PR is that a step should not fail quietly.

JJL772 (approved):

ruck314 (approved):
Both items from my previous review are addressed, and the pinned-header change resolves the extra-sync concern cleanly. I pushed the USAGE-line nit myself and trimmed the stale pageable-copy bullet and reorder paragraph from the description.

